PR #1, opened 2017-01-30
blocks: forked from examples-go/block_writer

petermattis:
Added support for mongodb and cassandra.

### DIFF
--- a/blocks/main.go
+++ b/blocks/main.go
@@ -18,6 +18,7 @@ package main
 import (
 	"bytes"
 	"database/sql"
+	"errors"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


blocks/main.go
```diff
@@ -0,0 +1,430 @@
...
+// existing table will be dropped.
+func setupDatabase(dbURL string) (database, error) {
+	parsedURL, err := url.Parse(dbURL)
+	if err != nil {
+		return nil, err
+	}
+	parsedURL.Path = "datablocks"
+
+	switch parsedURL.Scheme {
+	case "postgresql":
```
mberhault: we've been using `postgres` everywhere in the url. It's probably best to handle both here.

mberhault (approved):

blocks/main.go
petermattis: Done.

PR #2, opened 2017-01-30
blocks: support batch sizes >1 on cassandra

petermattis:


blocks/main.go
```diff
@@ -221,7 +220,7 @@ func setupMongo(parsedURL *url.URL) (database, error) {
 	}
 
 	session.SetMode(mgo.Monotonic, true)
-	// session.SetSafe(&mgo.Safe{FSync: true})
+	session.SetSafe(&mgo.Safe{FSync: true})
```
mberhault: I assume this is intentionally re-enabled.

mberhault (approved):

blocks/main.go
petermattis: Nope. Just experimentally. It destroy MongoDB performance. I'll comment it out again.

PR #3, opened 2017-01-30
blocks: use a hex-string instead of a placeholder for binary data

petermattis:
This improves throughput by 12% by reducing the round-trips to execute a
request. When using placeholders we have to prepare, then
bind+execute. Found when trying to figure out how cassandra and mongo
were executing requests faster than we could parse them (without
executing). This is an effort to make the comparison with
cassandra/mongo as apples-to-apples as possible.

Requires https://github.com/cockroachdb/cockroach/pull/13287

### DIFF
--- a/blocks/main.go
+++ b/blocks/main.go
@@ -134,20 +134,18 @@ type cockroach struct {
 
 func (c *cockroach) write(writerID string, blockNum int64, blockCount int, r *rand.Rand) error {
 	var buf bytes.Buffer
-	args := make([]interface{}, blockCount)
 	_, _ = buf.WriteString(
 		`INSERT INTO blocks (block_id, writer_id, block_num, raw_bytes) VALUES`)
 
 	for i := 0; i < blockCount; i++ {
 		blockID := r.Int63()
-		args[i] = randomBlock(r)
 		if i > 0 {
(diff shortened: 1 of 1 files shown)


bdarnell:
What if you use db.Prepare and Stmt.Exec (reusing the statement) instead of db.Exec(string)? That should be even faster than the inline literal.

petermattis:
> What if you use db.Prepare and Stmt.Exec (reusing the statement) instead of db.Exec(string)? That should be even faster than the inline literal.

With the `database/sql` interface, I believe that still has to prepare the statement on every connection. I can double-check later tonight.

bdarnell:
Yeah, you have to re-prepare on every connection, but connections are long-lived and the Stmt object knows how to handle connections that drop out from under it (according to http://go-database-sql.org/prepared.html). Also, I think this is a case of the go `lib/pq` being kind of dumb. The protocol doesn't require extra round trips; you can prepare/bind/execute all at once if you know the types of your arguments.

tamird:
+1 that if lib/pq is dumb then we should improve it - this speeds up our benchmark but our customers are likely to be using prepared statements.

petermattis:
Ok, see #4.

PR #9, opened 2017-02-02
kv: rename blocks to kv

petermattis:
Simplify schema to key/value. Rename "datablocks.blocks" to "test.kv".

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/9)


mberhault:
the makefile and build/ scripts need to be updated as well.

petermattis:
> the makefile and build/ scripts need to be updated as well.

Done.

kv/main.go
```diff
@@ -158,25 +152,25 @@ type cockroach struct {
...
 		args[j+0] = r.Int63()
-		args[j+1] = writerID
-		args[j+2] = blockNum + int64(i)
-		args[j+3] = randomBlock(r)
+		args[j+1] = randomBlock(r)
 	}
+	// TODO(peter): The key generation is not guaranteed unique. Consider using
+	// UPSERT, though initial tests show that is half the speed. Or perhaps
+	// ignoring duplicate key violation errors.
 	_, err := c.writeStmt.Exec(args...)
```
mberhault, line 174: this todo needs to be addressed, or runs without `--tolerate-errors` will fail sooner or later

mberhault (approved):

kv/main.go
petermattis: I agree it needs to be addressed, but are you saying it needs to be addressed now? 64-bit random numbers makes it pretty unlikely we'll hit this for short runs. For longer runs, --tolerate-errors seems adequate for this PR. I'm working on fixing this in a follow-on PR.

mberhault: Fair enough. We do have tolerate-errors everywhere anyway.

PR #11, opened 2017-02-02
Add some mongo sessions flags.

mberhault:
WMode and J only for now.
FSync means J when journaling is enabled, but we can't use both.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/11)


kv/main.go
```diff
@@ -64,6 +64,10 @@ var duration = flag.Duration("duration", 0, "The duration to run. If 0, run fore
 var benchmarkName = flag.String("benchmark-name", "BenchmarkBlocks", "Test name to report "+
 	"for Go benchmark results.")
 
+// Mongo flags. See https://godoc.org/gopkg.in/mgo.v2#Session.SetSafe for details.
+var mongoWMode = flag.String("mongo-wmode", "", "WMode for mongo session (eg: majority)")
```
petermattis, line 68: How many `wMode` options are there? Can you list the valid options in the help test (for the Mongo-newbies)?

petermattis (approved):
LGTM

kv/main.go
mberhault: it's the only one built-in, but people can define their own.
it used to be just `w: number of replicas`, but `majority` was added to make it independent of the replication factor.

PR #12, opened 2017-02-03
Add quorum/replication factor flags for cassandra.

mberhault:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/12)


petermattis (approved):
LGTM

mberhault:
my bad, these were both broken.
* replication: placeholders are not allowed in DDL statements. switched to generating the string
* Consistency.QUORUM is the default in gocql. Switched to a string parsed by gocql (quorum is still the default)

petermattis:
Still LGTM.

PR #13, opened 2017-02-03
kv: change key generation

petermattis:
Changed key generation to allow deterministic selection of keys for
reading that have already been written.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/13)

### DIFF
--- a/kv/main.go
+++ b/kv/main.go
@@ -17,9 +17,12 @@ package main
 
 import (
 	"bytes"
+	"crypto/sha1"
 	"database/sql"
+	"encoding/binary"
 	"flag"
 	"fmt"
+	"hash"
 	"log"
 	"math/rand"
 	"net/url"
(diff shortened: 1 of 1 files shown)


petermattis:
One downside to this change is that we generate the same sequence of keys to write on every invocation.

mberhault:
So this pretty much makes the `--drop` flag required?
I'm seeing kv failing on cassandra timeouts fairly regularly. This would make it tricky to resume a run. This is very preliminary though, we may be able to just change the retry options.

petermattis:
> So this pretty much makes the --drop flag required?

Any time you're starting writes, yes. Let me think about this some more to see if there is something else that can be done. One thought is to "seed" the sha1 hash with a random value but allow that seed to be specified manually as well.

petermattis:
Ok, I enhanced the key generation with a seed that can be manually specified but which defaults to `time.Now().UnixNano()`. So the default will be different keys on each run but you can specify `--seed` to make the keys deterministic between runs.

petermattis:
Ping.

kv/main.go
```diff
@@ -420,11 +491,12 @@ func main() {
 	var lastOps uint64
 	writers := make([]*blocker, *concurrency)
 
+	seq := &sequence{val: *writeSeq, seed: *seqSeed}
```
mberhault, line 494: can you log the initial and last (in `cast <-done`) sequence, that would make it easier to resume.

mberhault (approved):

kv/main.go
petermattis: Done.

mberhault:
LGTM (in case you missed the subtle "... approved these changes ..."

PR #16, opened 2017-02-08
Use one 'db' object per writer.

mberhault:
This is to guarantee separate sessions for mongo, but does not impact
other drivers.

TODO: surface errors so we can know if one or more writer never starts
on initialization

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/16)

### DIFF
--- a/kv/main.go
+++ b/kv/main.go
@@ -242,39 +242,38 @@ func (c *cockroach) write(count int, g *generator) error {
 	return err
 }
 
-func setupCockroach(parsedURL *url.URL) (database, error) {
+func setupCockroach(parsedURL *url.URL, firstTime bool) (database, error) {
 	// Open connection to server and create a database.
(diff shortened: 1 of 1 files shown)


petermattis:
Nice job tracking this down. While this works, I'd prefer not to muck with the `cassandra` and `cockroach` code. Let me think about whether there is a cleaner way to support this.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/16#-:-KcUWy5RS2ai0AA8mJO7:b-uevw9c)*

mberhault:
this was mostly a quick fix to check my assumptions, so not perfect. Although so far, my runs indicate that this is harmless in terms of performance. not a huge dataset though.

PR #17, opened 2017-02-08
kv: clone the mongo session

petermattis:
Apparently specifying "J: true" blocks all operations on an mgo.Session,
so provide a unique session per worker. I believe this is due to the
"monotonic" mode that we've specified for the session.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/17)

### DIFF
--- a/kv/main.go
+++ b/kv/main.go
@@ -162,6 +162,7 @@ func (g *generator) randomBlock() []byte {
 type database interface {
 	read(key int64) error
 	write(count int, g *generator) error
+	clone() database
 }
 
 type blocker struct {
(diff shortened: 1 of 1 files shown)


mberhault:
LGTM. you may want to call this `Copy`, there's already a `Clone` in `mgo.Session` which definitely doesn't do what we want.

petermattis:
> you may want to call this Copy, there's already a Clone in mgo.Session which definitely doesn't do what we want.

I intentionally went for `clone` because that's the terminology we use in the Cockroach code base.

PR #19, opened 2017-02-16
ycsb: specify full database.table

petermattis:
This avoids search path lookup and improves performance from 6,200
ops/sec to 14,700 ops/sec (130% !!!). Clearly we need to figure out why
the search path lookup is so costly.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/19)

### DIFF
--- a/ycsb/main.go
+++ b/ycsb/main.go
@@ -39,7 +39,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/cockroachdb/cockroach-go/crdb"
 	_ "github.com/lib/pq"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


petermattis:
@knz, @nvanbenschoten I'm going to take a look at why the search-path lookup is so costly. Just a heads up in case you know of something.

tamird:
LGTM (and wow)

a-robinson (approved):
WOW indeed

petermattis:
The problem is in `planner.QualifyWithDatabase`. If you don't specify the database in the query, that method performs a call to `searchAndQualifyDatabase` which is apparently very slow. I'm speculating that it is performing extra KV reads instead of looking in the cached `SystemConfig`. Or something. Regardless, almost certainly fixable. 

```
func (p *planner) QualifyWithDatabase(t *parser.NormalizableTableName) (*parser.TableName, error) {
	tn, err := t.Normalize()
	if err != nil {
		return nil, err
	}
	if tn.DatabaseName == "" {
		if err := p.searchAndQualifyDatabase(tn); err != nil {
			return nil, err
		}
	}
	return tn, nil
}
```

rjnn (approved):
What!? How did you even find this? LGTM.

petermattis:
> How did you even find this?

Profiling.

rjnn:
Sounds like we might benefit from a lunch and learn on profiling! :)

PR #20, opened 2017-02-17
ycsb: add --splits flag to pre-split the table

petermattis:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/20)

### DIFF
--- a/ycsb/main.go
+++ b/ycsb/main.go
@@ -73,6 +73,8 @@ var initialLoad = flag.Uint64("initial-load", 10000,
 var maxWrites = flag.Uint64("max-writes", 7*24*3600*1500,
 	"Maximum number of writes to perform before halting. This is required for accurately generating keys that are uniformly distributed across the keyspace.")
 
+var splits = flag.Int("splits", 0, "Number of splits to perform before starting normal operations")
+
 // ycsbWorker independently issues reads, writes, and scans against the database.
 type ycsbWorker struct {
 	workerID int
@@ -158,9 +160,11 @@ func newYcsbWorker(db *sql.DB, zipfR *ZipfGenerator, id int, workloadFlag string
 	}
 }
 
-func (yw *ycsbWorker) hashKey(key []byte, maxValue uint64) uint64 {
+func (yw *ycsbWorker) hashKey(key, maxValue uint64) uint64 {
+	yw.hashBuf = [8]byte{} // clear hashBuf
+	binary.PutUvarint(yw.hashBuf[:], key)
 	yw.hashFunc.Reset()
-	if _, err := yw.hashFunc.Write(key); err != nil {
+	if _, err := yw.hashFunc.Write(yw.hashBuf[:]); err != nil {
 		panic(err)
 	}
 	hashedKey := yw.hashFunc.Sum64()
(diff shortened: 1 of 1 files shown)


rjnn:
I'm fine with adding a splits flag, but I think in the long run this functionality should live in Cockroach itself - we should probably be rapidly pre-splitting ranges on cluster start until we at least have `k*numNodes` ranges for some single digit value of `k`.

---

Reviewed 1 of 1 files at r1.

---

*[ycsb/main.go, line 402 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/20#-Kd8mW_LGpMJDoQDjcKc:-Kd8mW_MsHd5KtlezDUc:b-3k9ea3) ([raw file](https://github.com/cockroachdb/loadgen/blob/10c1566274510d1465e32c7fc6dcccb85d92026c/ycsb/main.go#L402)):*
> ```Go
> 
> 	if *splits > 0 {
> 		r := rand.New(rand.NewSource(int64(time.Now().UnixNano())))
> ```

I'm not a fan of performing splits using a different random number generator, as it's not clear how split the table will actually be (what if you draw a bunch of splits that are in the middle of nowhere?). It's also not reproducible. How about using the same random number generator as the workers, and performing `n` splits after `hashFunc(1), hashFunc(2), ..., hashFunc(n)`? This guarantees that the hot keys from 1 through n are on different ranges.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/20#-:-Kd8n-G2huV52OYMws2r:b-e5b7qi)*

petermattis:
> I'm fine with adding a splits flag, but I think in the long run this functionality should live in Cockroach itself

Yes, Cockroach should automatically split based on load. But how do we know that will improve performance for this workload? By splitting manually and testing that performance improves. 

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycsb/main.go, line 402 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/20#-Kd8mW_LGpMJDoQDjcKc:-Kd8sKwmDEFWw2t4OwHm:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/10c1566274510d1465e32c7fc6dcccb85d92026c/ycsb/main.go#L402)):*

I'm not a fan of performing splits using a different random number generator, as it's not clear how split the table will actually be (what if you draw a bunch of splits that are in the middle of nowhere?). It's also not reproducible. How about using the same random number generator as the workers, and performing `n` splits after `hashFunc(1), hashFunc(2), ..., hashFunc(n)`? This guarantees that the hot keys from 1 through n are on different ranges.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/20#-:-Kd8qFqgwiUvXuCriObm:bau4u95)*

rjnn:
:lgtm:

Oh, yes, I agree. We definitely need this manual split flag as well. I just wanted to make sure we were on the same page with this making it into Cockroach as well. Apologies if that wasn't clear.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/20#-:-Kd8zQogfztX9NMWmWGm:bfjhrre)*

PR #21, opened 2017-02-17
ycsb: SetMax{Open,Idle}Conns like other load generators

petermattis:
This prevents any possibility of connection thrashing by database/sql.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/21)


rjnn:
LGTM.

PR #28, opened 2017-03-06
kv: remove unnecessary LIMIT

petermattis:
We're selecting on the primary key, so there will always be at most one
result.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/28)


bdarnell:
LGTM. Does this change performance or is it just a cleanup?

petermattis:
No performance change. Just a cleanup.

PR #30, opened 2017-03-08
ycsb: fix --initial-load

petermattis:
Fix ycsbWorker.runLoader so that it actually loads all of the keys
specified. The previous code was dividing two integers which resulted in
precision loss if the values didn't divide evenly. Additionally, there
was an off-by-one error in the calculation of the first key to add
because zipfIMin is set to 1, not 0.

Relax the restriction that IMin must be less than IMax. It can be equal
as well without problem. This allows --initial-load=1 to
work (i.e. write a single key).

Together, these changes allow `--initial-load=1 --workload=C` to
generate a 100% read workload to a single key for any concurrency
setting.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/30)


rjnn (approved):
LGTM

PR #31, opened 2017-03-08
ycsb: protect zetaN with the mutex

petermattis:
ZipfGenerator.Uint64 was accessing ZipfGenerator.zetaN without a lock
even though this field is updated concurrently by
ZipfGenerator.IncrementIMax.

Fix buglet in --initial-load. We were loading 1 too many keys initially,
causing the first write to hit a duplicate key constraint error.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/31)

### DIFF
--- a/ycsb/main.go
+++ b/ycsb/main.go
@@ -77,8 +77,7 @@ var splits = flag.Int("splits", 0, "Number of splits to perform before starting 
 
 // ycsbWorker independently issues reads, writes, and scans against the database.
 type ycsbWorker struct {
-	workerID int
-	db       *sql.DB
+	db *sql.DB
 	// An RNG used to generate random keys
 	zipfR *ZipfGenerator
 	// An RNG used to generate random strings for the values
@@ -114,10 +113,7 @@ const (
 	scanOp
 )
 
-func newYcsbWorker(db *sql.DB, zipfR *ZipfGenerator, id int, workloadFlag string) *ycsbWorker {
-	if *verbose {
-		fmt.Printf("Creating YCSB Worker '%d' running Workload %s\n", id, workloadFlag)
-	}
+func newYcsbWorker(db *sql.DB, zipfR *ZipfGenerator, workloadFlag string) *ycsbWorker {
 	source := rand.NewSource(int64(time.Now().UnixNano()))
 	var readFreq, writeFreq, scanFreq float32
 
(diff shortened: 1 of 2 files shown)


petermattis:
@a-robinson I think @arjunravinarayan is on vacation. Noticed this during code inspection. I'm not sure how it could cause the `denim` perf oddities, but worth fixing regardless.

ycsb/main.go
```diff
@@ -206,7 +206,7 @@ func (yw *ycsbWorker) runLoader(n uint64, numWorkers int, thisWorkerNum int, wg
 	if *verbose {
 		fmt.Printf("Worker %d loading %d rows of data\n", yw.workerID, n)
```
a-robinson: Not your code, but this statement appears incorrect - worker `yw.workerID` is loading much less than `n` rows of data judging by the loop below

a-robinson (approved):
LGTM

ycsb/main.go
petermattis: I've removed this log message and a few others which weren't particularly useful.

petermattis:
Updated to move `ZipfGenerator.r` under the mutex. PTAL.

ycsb/main.go
```diff
@@ -466,9 +448,12 @@ func main() {
 
 	loadStart := time.Now()
 	var wg sync.WaitGroup
-	for i := range workers {
+	// TODO(peter): Using all of the workers for loading leads to errors with
+	// some of the insert statements receiving an EOF. For now, use a single
+	// worker.
+	for i, n := 0, 1; i < n; i++ {
```
a-robinson, line 454: It's fine if we actually plan on coming back and changing it, but this is a really strange way of doing something once.

a-robinson (approved):
LGTM. Did this fix things?

ycsb/main.go
petermattis: I'm going to take a look at this tonight.

petermattis:
> Did this fix things?

Yes. See the latest message in https://github.com/cockroachdb/cockroach/issues/13627.

rjnn:
Oof, good catch.

---

Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/31#-:-Kf7K8MqHNJcjFfI3tDj:b199fy9)*

PR #35, opened 2017-03-15
ycsb: fix ycsb test failure

rjnn:
Seed as a variable was removed in #31, but not from this test, breaking the test.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/35)


tamird:
LGTM

PR #38, opened 2017-03-17
kv: remove p50 latency reporting

petermattis:
The p50 latency isn't terribly interesting and can create a sense of
false optimism in performance. p95, p99 and pMax are much more valuable.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/38)


mberhault:
LGTM

PR #45, opened 2017-04-13
kv: track and output cumulative latency

petermattis:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/45)

### DIFF
--- a/kv/main.go
+++ b/kv/main.go
@@ -560,6 +560,8 @@ func main() {
 			*benchmarkName, numOps, float64(elapsed.Nanoseconds())/float64(numOps))
 	}()
 
+	cumLatency := hdrhistogram.New(minLatency.Nanoseconds(), maxLatency.Nanoseconds(), 1)
+
 	for i := 0; ; {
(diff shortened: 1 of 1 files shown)


kv/main.go
```diff
@@ -585,6 +587,7 @@ func main() {
 				}
 			}
 
+			cumLatency.Merge(h)
```
rjnn, line 590: Merge returns a value (number of values "dropped"). Do you know what that is and if we should care?

rjnn (approved):
LGTM

kv/main.go
petermattis: It returns the number of samples that could not be merged. I believe that can only happen if you merge 2 histograms that were configured differently. Our histograms are all configured identically.

rjnn: Great. I could not figure that out from the docs. Thanks.

petermattis: Docs lie. Read the source.

PR #51, opened 2017-05-09
tpch: continue on error

rjnn:
We want continuous load tests to continue even if queries return
errors, unless the max error count is encountered.

Also fixes the fact that this return path was bypassing the
WaitGroup. Moved the Done() to a defer in order to prevent errors of
this type from reoccuring.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/loadgen/51)

### DIFF
--- a/tpch/main.go
+++ b/tpch/main.go
@@ -30,13 +30,13 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/cockroachdb/cockroach-go/crdb"
 	_ "github.com/lib/pq"
 	"github.com/pkg/errors"
-	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


cuongdo:
:lgtm:

In the future, it might be good to have a `--tolerate-errors` flag like some of our other load generators have. Sometimes, we may want strictness.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/51#-:-KjiyRdolGVXp44Cz8MN:b8bd9np)*

rjnn:
We do have a `--max-errors` flag, which will error out on the `Fatal` code path already. Are you suggesting we rename the flag to fit the pattern, or something else?

PR #52, opened 2017-06-27
ycsb: fix duplicate key inserts

petermattis:
Do not mod the hashed key by --max-writes. Doing so limited the key
space dramatically and resulted in an increasing number of collisions as
we got closer to the --max-writes value.

Add --write-duration flag which controls how long the test performs
writes. After the duration has elapsed the load generator reverts to
only performing reads. This is useful during performance investigation
to determine how reads perform after a large amount of data has been
inserted.

### DIFF
--- a/ycsb/main.go
+++ b/ycsb/main.go
@@ -28,6 +28,7 @@ import (
 	"hash"
 	"hash/fnv"
 	"log"
+	"math"
 	"math/rand"
 	"net/url"
 	"os"
@@ -59,11 +60,13 @@ const (
 
 var concurrency = flag.Int("concurrency", 2*runtime.NumCPU(),
 	"Number of concurrent workers sending read/write requests.")
-var workload = flag.String("workload", "B", "workload type. Choose from A-E.")
+var workload = flag.String("workload", "B", "workload type. Choose from A-F.")
 var tolerateErrors = flag.Bool("tolerate-errors", false,
 	"Keep running on error. (default false)")
 var duration = flag.Duration("duration", 0,
 	"The duration to run. If 0, run forever.")
+var writeDuration = flag.Duration("write-duration", 0,
+	"The duration to perform writes. If 0, write forever.")
 var verbose = flag.Bool("v", false, "Print *verbose debug output")
 var drop = flag.Bool("drop", true,
 	"Drop the existing table and recreate it to start from scratch")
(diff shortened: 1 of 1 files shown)


rjnn:
:lgtm: 

I reviewed the YCSB paper, and this fix checks out. I'm a little puzzled that I originally interpreted the mod requirement as a mod by `max-writes`. This is a clear improvement, and more in line with the spec. I'm curious: what investigation sent you down this path?

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[ycsb/main.go, line 63 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/52#-KnezfAMftmh9jrkfKYm:-KnezfAMftmh9jrkfKYn:b-m2nuj9) ([raw file](https://github.com/cockroachdb/loadgen/blob/172dd568c58c666537ca973c8673a86eb4dde0ea/ycsb/main.go#L63)):*
> ```Go
> var concurrency = flag.Int("concurrency", 2*runtime.NumCPU(),
> 	"Number of concurrent workers sending read/write requests.")
> var workload = flag.String("workload", "B", "workload type. Choose from A-E.")
> ```

s/A-E/A-F/ with your new addition.

---

*[ycsb/main.go, line 189 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/52#-KnezpFrieiXGw4MqIQ6:-KnezpFrieiXGw4MqIQ7:b5v0q8u) ([raw file](https://github.com/cockroachdb/loadgen/blob/172dd568c58c666537ca973c8673a86eb4dde0ea/ycsb/main.go#L189)):*
> ```Go
> 		panic(err)
> 	}
> 	// The Go sql driver interface does not support having the high-bit set in
> ```

What did you do that led you to find this out?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/52#-:-KnezDeY8YF-_h0wYqx1:b-jd96xq)*

petermattis:
> I'm curious: what investigation sent you down this path?

I noticed a large and increasing number of aborts when running `ycsb`. Finally decided to track down the cause.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycsb/main.go, line 63 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/52#-KnezfAMftmh9jrkfKYm:-Knf1AN1t4_-klM9kjF7:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/172dd568c58c666537ca973c8673a86eb4dde0ea/ycsb/main.go#L63)):*

s/A-E/A-F/ with your new addition.
</blockquote></details>

Done.

---

*[ycsb/main.go, line 189 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/52#-KnezpFrieiXGw4MqIQ6:-Knf1BK71qmKlGVauMe8:b-rwpfz2) ([raw file](https://github.com/cockroachdb/loadgen/blob/172dd568c58c666537ca973c8673a86eb4dde0ea/ycsb/main.go#L189)):*

What did you do that led you to find this out?
</blockquote></details>

I originally didn't have it and quickly hit an error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/52#-:-Knf1QpWH0aHVuDSWOAV:b-fkow66)*

PR #56, opened 2017-07-19
kv: Add option to generate keys sequentially

a-robinson:
This could also easily be tweaked to use a separate starting point for each concurrent writer if we want to test that workload as well. For now, all writers use the same sequence (as for the hash generator).

### DIFF
--- a/kv/main.go
+++ b/kv/main.go
@@ -73,6 +73,7 @@ var maxOps = flag.Uint64("max-ops", 0, "Maximum number of blocks to read/write")
 var duration = flag.Duration("duration", 0, "The duration to run. If 0, run forever.")
 var writeSeq = flag.Int64("write-seq", 0, "Initial write sequence value.")
 var seqSeed = flag.Int64("seed", time.Now().UnixNano(), "Key hash seed.")
+var sequential = flag.Bool("sequential", false, "Pick keys sequentially instead of randomly.")
 var drop = flag.Bool("drop", false, "Clear the existing data before starting.")
 var benchmarkName = flag.String("benchmark-name", "BenchmarkBlocks", "Test name to report "+
 	"for Go benchmark results.")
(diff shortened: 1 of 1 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[kv/main.go, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/56#-KpROkR9DqBM67-YdJww:-KpROkR9DqBM67-YdJwx:btf64z8) ([raw file](https://github.com/cockroachdb/loadgen/blob/9f30184a884815d0ac18bcdb6ffbe5818a35ff42/kv/main.go#L193)):*
> ```Go
> 		return oldest
> 	}
> 	return oldest + g.random.Int63n(newest-oldest)
> ```

The expectation is that `--write-seq` can be specified to pick up where a previous run left off. That means that you should be able to read values from `[0, g.seq.read()]`. I'm not seeing why the implementation of `readKey` here is different from `hashGenerator.readKey`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/56#-:-KpRP3Oltb_IEYxm7K2J:bnfp4nl)*

a-robinson:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[kv/main.go, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/56#-KpROkR9DqBM67-YdJww:-KpRPDcb-NJGshlNX0KQ:bed2v6g) ([raw file](https://github.com/cockroachdb/loadgen/blob/9f30184a884815d0ac18bcdb6ffbe5818a35ff42/kv/main.go#L193)):*

The expectation is that `--write-seq` can be specified to pick up where a previous run left off. That means that you should be able to read values from `[0, g.seq.read()]`. I'm not seeing why the implementation of `readKey` here is different from `hashGenerator.readKey`.
</blockquote></details>

Ah, ok. It'll break reads if someone specifies a non-zero `writeSeq` without having previously run this, but I now see that was already the case. I've changed it back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/56)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[kv/main.go, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/56#-KpROkR9DqBM67-YdJww:-KpRQy80qWJp487zdACn:bb77ucv) ([raw file](https://github.com/cockroachdb/loadgen/blob/9f30184a884815d0ac18bcdb6ffbe5818a35ff42/kv/main.go#L193)):*
<details><summary><i>Previously, a-robinson (Alex Robinson) wrote…</i></summary><blockquote>

Ah, ok. It'll break reads if someone specifies a non-zero `writeSeq` without having previously run this, but I now see that was already the case. I've changed it back.
</blockquote></details>

Yep.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/56)*

PR #60, opened 2017-07-25
ycsb: add strict postgres mode

rjnn:
The ycsb load generator doesn't work when run against postgres because
we've added various cockroach specific features.

It doesn't make sense to add a new worker type because we set the
worker type based on URL scheme, so add a "strict" mode. This still
works with CockroachDB, albeit with reduced performance.

### DIFF
--- a/ycsb/main.go
+++ b/ycsb/main.go
@@ -75,6 +75,8 @@ var rateLimit = flag.Uint64("rate-limit", 0,
 	"Maximum number of operations per second per worker. Set to zero for no rate limit")
 var initialLoad = flag.Uint64("initial-load", 10000,
 	"Initial number of rows to sequentially insert before beginning Zipfian workload generation")
+var strictPostgres = flag.Bool("strict-postgres", false,
+	"Use Postgres compatible syntax, without any Cockroach specific extensions")
 
 // 7 days at 5% writes and 30k ops/s
 var maxWrites = flag.Uint64("max-writes", 7*24*3600*1500,
@@ -399,10 +401,22 @@ func setupCockroach(parsedURL *url.URL) (database, error) {
 		}
 	}
 
+	if *strictPostgres {
+		// Since we use absolute paths (ycsb.usertable), create a Postgres schema
+		// to make the absolute paths work.
+		if _, err := db.Exec("CREATE SCHEMA IF NOT EXISTS ycsb"); err != nil {
+			if *verbose {
+				fmt.Printf("Failed to create schema: %s\n", err)
+			}
+			// Do not fail on this error, as we want strict postgres mode to
+			// also work on Cockroach, and Cockroach doesn't have CREATE SCHEMA.
+		}
+	}
+
 	// Create the initial table for storing blocks.
 	createStmt := `
 CREATE TABLE IF NOT EXISTS ycsb.usertable (
-	ycsb_key INT PRIMARY KEY NOT NULL,
+	ycsb_key BIGINT PRIMARY KEY NOT NULL,
 	FIELD1 TEXT,
 	FIELD2 TEXT,
 	FIELD3 TEXT,


tamird:
Does this actually reduce performance? I'm confused as to why our search path stuff takes a hit when the database is unspecified. @knz?

petermattis:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycsb/main.go, line 342 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUJWfZs_JNcvWPyuv:-KpuUJWfZs_JNcvWPyuw:big0hzb) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L342)):*
> ```Go
> 	var readString string
> 	if *strictPostgres {
> 		readString = fmt.Sprintf("SELECT * FROM usertable WHERE ycsb_key=%d", key)
> ```

Feels like there is some schema path that could be set to allow the cockroach syntax to work on postgres.

---

*[ycsb/main.go, line 396 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUQkpD7dKN2cuKfG3:-KpuUQkqA9ihQuEZtSwt:b50xwey) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L396)):*
> ```Go
> 
> 	if *strictPostgres {
> 		fmt.Println("In strict postgres mode, assuming database exists. Use 'createdb' to manually create")
> ```

I think you can do `CREATE DATABASE ycsb` and ignore the error.

---

*[ycsb/main.go, line 412 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUZYIfSQbg6xwPCxQ:-KpuUZYIfSQbg6xwPCxR:b-g7cd9c) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L412)):*
> ```Go
> 		var dropString string
> 		if *strictPostgres {
> 			dropString = "DROP TABLE IF EXISTS usertable"
> ```

This isn't performance critical. Should be fine to allow uses this statement.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-:-KpuUh7nULUMH7aYhVv8:bnfp4nl)*

rjnn:
`ALTER TABLE SPLIT`, `SCATTER`, etc. are CockroachDB specific performance enhancing things that need to be disabled anyway. At one point we also had bad performance implications when not specifying absolute paths, I'm not sure what the status of that is now.

---

Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycsb/main.go, line 342 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUJWfZs_JNcvWPyuv:-KpuZllJX_VYAxl2_d_c:b-h7lxxf) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L342)):*

Feels like there is some schema path that could be set to allow the cockroach syntax to work on postgres.
</blockquote></details>

Good idea. There is (create a schema named "ycsb" under database "ycsb", with table "usertable"). Done.

---

*[ycsb/main.go, line 396 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUQkpD7dKN2cuKfG3:-Kpu_BLUuBZDlbG_vwoP:b-yn3llo) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L396)):*

I think you can do `CREATE DATABASE ycsb` and ignore the error.
</blockquote></details>

True, in fact we're already ignoring the error... Removed.

---

*[ycsb/main.go, line 412 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUZYIfSQbg6xwPCxQ:-Kpu_L_39mMAUvVIZu2v:b-h0r36f) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L412)):*

This isn't performance critical. Should be fine to allow uses this statement.
</blockquote></details>

Now that we're using a schema named `ycsb`, retaining `ycsb.usertable` in the drop statement.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-:-Kpu_Xx-etTzx5yFAFEu:b-4mrek5)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycsb/main.go, line 396 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUQkpD7dKN2cuKfG3:-KpulLBK3aTnWAY5FpVk:b9eycti) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L396)):*

True, in fact we're already ignoring the error... Removed.
</blockquote></details>

Don't you have to get rid of the `IF NOT EXISTS` below? That syntax isn't supported by Postgres, right?

---

*[ycsb/main.go, line 414 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpulWxLkTbLHHU05CTs:-KpulWxLkTbLHHU05CTt:blu1tmh) ([raw file](https://github.com/cockroachdb/loadgen/blob/34b2ddbb556b19671c5d5cb2d5d73d0ccf1bf6c6/ycsb/main.go#L414)):*
> ```Go
> 			// also work on Cockroach, and Cockroach doesn't have CREATE SCHEMA.
> 		}
> 	}
> ```

I would think this belongs up near the `CREATE DATABASE` statement.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/60)*

rjnn:
*[ycsb/main.go, line 396 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpuUQkpD7dKN2cuKfG3:-KpunYmI04nHaKlRlz8C:b6obcnd) ([raw file](https://github.com/cockroachdb/loadgen/blob/f0e679888c61e62dbac17ad708c070882afbdc7d/ycsb/main.go#L396)):*
```
> psql
ycsb=# create table ycsb.foo (i int);
CREATE TABLE
ycsb=# create table if not exists ycsb.foo (i int);
NOTICE:  relation "foo" already exists, skipping
CREATE TABLE
ycsb=# create table if not exists ycsb.bar (i int);
CREATE TABLE
```

postgres doesn't support `create database` or `create database if not exists`. It supports `create table if not exists`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/60)*

rjnn:
*[ycsb/main.go, line 414 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/60#-KpulWxLkTbLHHU05CTs:-Kpunr3MYx35h1udI9ws:b-88bvjv) ([raw file](https://github.com/cockroachdb/loadgen/blob/34b2ddbb556b19671c5d5cb2d5d73d0ccf1bf6c6/ycsb/main.go#L414)):*

I would think this belongs up near the `CREATE DATABASE` statement.
</blockquote></details>

Fair.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/60)*

PR #63, opened 2017-08-04
tpcc: implement data loader

jordanlewis:
Add implementation of the data load component of TPCC.

See http://www.tpc.org/tpc_documents_current_versions/pdf/tpc-c_v5.11.0.pdf

### DIFF
--- a/tpcc/ddls.sql
+++ b/tpcc/ddls.sql
@@ -1,0 +1,144 @@
+-- TODO(jordan): put "if not exists" behind a flag since it's non-standard
+-- SQL - we'd like to be able to run this against pg as well.
+create database if not exists tpcc;
+
+create table if not exists warehouse (
+  w_id        integer   not null primary key,
+  w_name      varchar(10),
+  w_street_1  varchar(20),
+  w_street_2  varchar(20),
+  w_city      varchar(20),
+  w_state     char(2),
+  w_zip       char(9),
+  w_tax       decimal(4,4),
+  w_ytd       decimal(12,2)
+);
+
+-- 10 districts per warehouse
+create table if not exists district (
+  d_id         integer       not null,
+  d_w_id       integer       not null,
+  d_name       varchar(10),
+  d_street_1   varchar(20),
+  d_street_2   varchar(20),
+  d_city       varchar(20),
+  d_state      char(2),
+  d_zip        char(9),
+  d_tax        decimal(4,4),
+  d_ytd        decimal(12,2),
+  d_next_o_id  integer,
+  primary key (d_w_id, d_id),
+  foreign key (d_w_id) references warehouse (w_id)
+);
+
+create table if not exists customer (
+  c_id           integer        not null,
+  c_d_id         integer        not null,
+  c_w_id         integer        not null,
+  c_first        varchar(16),
+  c_middle       char(2),
+  c_last         varchar(16),
+  c_street_1     varchar(20),
+  c_street_2     varchar(20),
+  c_city         varchar(20),
+  c_state        char(2),
+  c_zip          char(9),
+  c_phone        char(16),
+  c_since        timestamp,
+  c_credit       char(2),
+  c_credit_lim   decimal(12,2),
+  c_discount     decimal(4,4),
+  c_balance      decimal(12,2),
+  c_ytd_payment  decimal(12,2),
+  c_payment_cnt  integer,
+  c_delivery_cnt integer,
+  c_data         varchar(500),
+  primary key (c_w_id, c_d_id, c_id),
+  foreign key (c_w_id, c_d_id) references district (d_w_id, d_id)
+);
+
+-- No PK necessary for this table.
+create table if not exists history (
+  h_c_id   integer,
+  h_c_d_id integer,
+  h_c_w_id integer,
+  h_d_id   integer,
+  h_w_id   integer,
+  h_date   timestamp,
+  h_amount decimal(6,2),
+  h_data   varchar(24),
+  foreign key (h_c_w_id, h_c_d_id, h_c_id) references customer (c_w_id, c_d_id, c_id),
+  foreign key (h_w_id, h_d_id) references district (d_w_id, d_id)
+);
+
+create table if not exists "order" (
+  o_id         integer      not null,
+  o_d_id       integer      not null,
+  o_w_id       integer      not null,
+  o_c_id       integer,
+  o_entry_d    timestamp,
+  o_carrier_id integer,
+  o_ol_cnt     integer,
+  o_all_local  integer,
+  primary key (o_w_id, o_d_id, o_id),
+  foreign key (o_w_id, o_d_id, o_c_id) references customer (c_w_id, c_d_id, c_id)
+);
+
+create table if not exists new_order (
+  no_o_id  integer   not null,
+  no_d_id  integer   not null,
+  no_w_id  integer   not null,
+  primary key (no_w_id, no_d_id, no_o_id),
+  foreign key (no_w_id, no_d_id, no_o_id) references "order" (o_w_id, o_d_id, o_id)
+);
+
+create table if not exists item (
+  i_id     integer      not null,
+  i_im_id  integer,
+  i_name   varchar(24),
+  i_price  decimal(5,2),
+  i_data   varchar(50),
+  primary key (i_id)
+);
+
+create table if not exists stock (
+  s_i_id       integer       not null,
+  s_w_id       integer       not null,
+  s_quantity   integer,
+  s_dist_01    char(24),
+  s_dist_02    char(24),
+  s_dist_03    char(24),
+  s_dist_04    char(24),
+  s_dist_05    char(24),
+  s_dist_06    char(24),
+  s_dist_07    char(24),
+  s_dist_08    char(24),
+  s_dist_09    char(24),
+  s_dist_10    char(24),
+  s_ytd        integer,
+  s_order_cnt  integer,
+  s_remote_cnt integer,
+  s_data       varchar(50),
+  primary key (s_w_id, s_i_id),
+  foreign key (s_w_id) references warehouse (w_id),
+  foreign key (s_i_id) references item (i_id)
+);
+
+create table if not exists order_line (
+  ol_o_id         integer   not null,
+  ol_d_id         integer   not null,
+  ol_w_id         integer   not null,
+  ol_number       integer   not null,
+  ol_i_id         integer   not null,
+  ol_supply_w_id  integer,
+  ol_delivery_d   timestamp,
+  ol_quantity     integer,
+  ol_amount       decimal(6,2),
+  ol_dist_info    char(24),
+  primary key (ol_w_id, ol_d_id, ol_o_id, ol_number),
+  foreign key (ol_w_id, ol_d_id, ol_o_id) references "order" (o_w_id, o_d_id, o_id),
+  foreign key (ol_supply_w_id, ol_i_id) references stock (s_w_id, s_i_id)
+);
+
+create index if not exists customer_idx on customer (c_w_id, c_d_id, c_last, c_first);
+create unique index if not exists order_idx on "order" (o_w_id, o_d_id, o_carrier_id, o_id);
--- a/tpcc/generate.go
+++ b/tpcc/generate.go
@@ -1,0 +1,296 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License. See the AUTHORS file
+// for names of contributors.
+
+package main
+
+import (
+	"database/sql"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"strings"
+	"sync"
+	"time"
+)
+
+// These constants are all set by the spec - they're not knobs. Don't change
+// them.
+const nItems = 100000
+const nStock = 100000
+const nCustomers = 3000
+
+// District constants
+const ytd = 30000.00 // also used by warehouse
+const nextOrderId = 3001
+
+// Customer constants
+const creditLimit = 50000.00
+const balance = -10.00
+const ytdPayment = 10.00
+const middleName = "OE"
+const paymentCount = 1
+const deliveryCount = 0
+const goodCredit = "GC"
+const badCredit = "BC"
+
+// loadSchema loads the entire TPCC schema into the database.
+func loadSchema(db *sql.DB) {
+	data, err := ioutil.ReadFile("ddls.sql")
+	if err != nil {
+		panic(err)
+	}
+	stmts := strings.Split(string(data), ";")
+	for _, stmt := range stmts {
+		if _, err := db.Exec(stmt); err != nil {
+			panic(err)
+		}
+	}
+}
+
+func prepare(db *sql.DB, query string) *sql.Stmt {
+	stmt, err := db.Prepare(query)
+	if err != nil {
+		panic(err)
+	}
+	return stmt
+}
+
+func parallelLoad(n int, entityName string, loader func(int)) {
+	var wg sync.WaitGroup
+	ch := make(chan int)
+	for i := 0; i < *concurrency; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for id := range ch {
+				loader(id)
+			}
+		}()
+	}
+
+	for id := 1; id <= n; id++ {
+		ch <- id
+		if id%1000 == 0 {
+			fmt.Printf("Loaded %d/%d %ss\n", id, n, entityName)
+		}
+	}
+	close(ch)
+	wg.Wait()
+}
+
+func generateData(db *sql.DB) {
+	stmtItem := prepare(db, `
+INSERT INTO item (i_id, i_im_id, i_name, i_price, i_data)
+VALUES ($1, $2, $3, $4, $5)`)
+	stmtWarehouse := prepare(db, `
+INSERT INTO warehouse (w_id, w_name, w_street_1, w_street_2, w_city, w_state, w_zip, w_tax, w_ytd)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)`)
+	stmtStock := prepare(db, `
+INSERT INTO stock (
+	s_i_id, s_w_id, s_quantity,
+	s_dist_01, s_dist_02, s_dist_03, s_dist_04, s_dist_05, s_dist_06, s_dist_07, s_dist_08, s_dist_09, s_dist_10,
+	s_ytd, s_order_cnt, s_remote_cnt, s_data)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17)`)
+	stmtDistrict := prepare(db, `
+INSERT INTO district (
+	d_id, d_w_id, d_name, d_street_1, d_street_2,
+	d_city, d_state, d_zip, d_tax, d_ytd, d_next_o_id)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11)`)
+	stmtCustomer := prepare(db, `
+INSERT INTO customer (
+	c_id, c_d_id, c_w_id, c_first, c_middle, c_last,
+	c_street_1, c_street_2, c_city, c_state, c_zip,
+	c_phone, c_since, c_credit, c_credit_lim, c_discount,
+	c_balance, c_ytd_payment, c_payment_cnt,
+	c_delivery_cnt, c_data)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21)`)
+	stmtHistory := prepare(db, `
+INSERT INTO history (h_c_id, h_c_d_id, h_c_w_id, h_d_id, h_w_id, h_date, h_amount, h_data)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8)`)
+	stmtOrder := prepare(db, `
+INSERT INTO "order" (o_id, o_d_id, o_w_id, o_c_id, o_entry_d, o_carrier_id, o_ol_cnt, o_all_local)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8)`)
+	stmtOrderLine := prepare(db, `
+INSERT INTO order_line (ol_o_id, ol_d_id, ol_w_id, ol_number, ol_i_id, ol_supply_w_id, ol_delivery_d, ol_quantity, ol_amount, ol_dist_info)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)`)
+	stmtNewOrder := prepare(db, `
+INSERT INTO new_order (no_o_id, no_d_id, no_w_id)
+VALUES ($1, $2, $3)`)
+
+	// See section 1.3 for the general layout of the tables.
+	// See section 4.3 for the rules on how to populate the database.
+
+	fmt.Println("Loading items...")
+
+	// 100,000 items.
+	// TODO(jordan): send these inserts in batches.
+	parallelLoad(nItems, "item", func(id int) {
+		if _, err := stmtItem.Exec(id,
+			randInt(1, 10000),                         // im_id: "Image ID associated to Item"
+			randAString(14, 24),                       // name
+			float64(randInt(100, 10000))/float64(100), // price
+			randOriginalString(),
+		); err != nil {
+			panic(err)
+		}
+	})
+
+	now := time.Now()
+	for w_id := 0; w_id < *warehouses; w_id++ {
+		fmt.Printf("Loading warehouse %d...\n", w_id)
+		stmtWarehouse.Exec(w_id,
+			randInt(6, 10),  // name
+			randInt(10, 20), // street_1
+			randInt(10, 20), // street_2
+			randInt(10, 20), // city
+			randState(),
+			randZip(),
+			randTax(),
+			ytd)
+
+		// 100,000 stock per warehouse.
+		parallelLoad(nStock, "stock", func(id int) {
+			if _, err := stmtStock.Exec(id, w_id,
+				randInt(10, 100),     // quantity
+				randAString(24, 24),  // dist_01
+				randAString(24, 24),  // dist_02
+				randAString(24, 24),  // dist_03
+				randAString(24, 24),  // dist_04
+				randAString(24, 24),  // dist_05
+				randAString(24, 24),  // dist_06
+				randAString(24, 24),  // dist_07
+				randAString(24, 24),  // dist_08
+				randAString(24, 24),  // dist_09
+				randAString(24, 24),  // dist_10
+				0,                    // ytd
+				0,                    // order_cnt
+				0,                    // remote_cnt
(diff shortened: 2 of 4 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[tpcc/generate.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqoy-xHYVjQI4R1qYE1:-Kqoy-xHYVjQI4R1qYE2:b-vm05ve) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L30)):*
> ```Go
> )
> 
> // These constants are all sent by the spec - they're not knobs. Don't change
> ```

`s/sent/set/g`

---

*[tpcc/generate.go, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp054tp7T9TJkQsVS1:-Kqp054tp7T9TJkQsVS2:b3ontkk) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L58)):*
> ```Go
> 	stmts := strings.Split(string(data), ";")
> 	for _, stmt := range stmts {
> 		if err := crdb.ExecuteTx(db, func(tx *sql.Tx) error {
> ```

With only a single statement, there isn't much reason to be using a transaction vs executing the statement in an implicit transaction.

---

*[tpcc/generate.go, line 75 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqozFeckv1KlZFulnjP:-KqozFeckv1KlZFulnjQ:bkpwra5) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L75)):*
> ```Go
> }
> 
> func startWorkerPool(loader func(int), n int, entityName string) {
> ```

This name implies a function which returns a worker pool (to me, at least). But really what it is does run `loader` `n` times with `*concurrency` workers. Perhaps rename to `parallel` or something like that. Also, small nit: I think calling this function looks better if the function is last. Consider:

```
startWorkerPool(n, "foo", func(int) {
  // load the foos
})
```

vs 

```
startWorkerPool(func(int) {
  // load the foos
}, n "foo")
```

---

*[tpcc/generate.go, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqoxYPAowB5y9duCKz7:-KqoxYPAowB5y9duCKz8:by71607) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L99)):*
> ```Go
> 
> func generateData(db *sql.DB) {
> 	stmtItem := prepare(db, `INSERT INTO item (i_id, i_im_id, i_name, i_price, i_data) VALUES ($1, $2, $3, $4, $5)`)
> ```

I think you can use a little bit more formatting in these statements to make them more readable. For example:

```
stmtItem := prepare(db, `
INSERT INTO item (i_id, i_im_id, i_name, i_price, i_data) 
  VALUES ($1, $2, $3, $4, $5)
`)
```

---

*[tpcc/generate.go, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0iDQS-fyLYdcIktt:-Kqp0iDQS-fyLYdcIktu:b-g0fysu) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L139)):*
> ```Go
> 			ytd)
> 
> 		// 100,000 stock per warehouse..
> ```

`s/.././g`

---

*[tpcc/generate.go, line 141 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0YZUEm7VJHYcVsMY:-Kqp0YZUEm7VJHYcVsMZ:bvygl0f) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L141)):*
> ```Go
> 		// 100,000 stock per warehouse..
> 		startWorkerPool(func(id int) {
> 			if _, err := stmtStock.Exec(id, w_id,
> ```

Does the spec say anything about how loading occurs? We'd likely see faster loading by batching inserts. If this is kosher, add a `TODO`.

---

*[tpcc/main.go, line 44 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp-NxF44ec805EKo3Y:-Kqp-NxF44ec805EKo3Z:bylh67) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/main.go#L44)):*
> ```Go
> func setupDatabase(dbURL string) (*sql.DB, error) {
> 	if *verbose {
> 		log.Printf("connecting to db: %s\n", dbURL)
> ```

IMO, `log.Printf` is overly verbose due to the prefixed timestamp info. For these load generators, I prefer `fmt.Printf` and including elapsed times when appropriate.

---

*[tpcc/random.go, line 94 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqmm0wiPtH9iUBpsISy:-Kqmm0wiPtH9iUBpsISz:b-kxwg2z) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L94)):*
> ```Go
> // randTax produces a random tax between [0.0000..0.2000]
> func randTax() float64 {
> 	return float64(randInt(0, 2000)) / float64(10000.0)
> ```

Is this specified? I would have though you'd do `rand.Float64() * 0.2`, though I guess that isn't inclusive at the end range.

---

*[tpcc/random.go, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqmmW0uqhCoEZiCcrN8:-KqmmW0uqhCoEZiCcrN9:b-3oao8v) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L99)):*
> ```Go
> // randInt returns a number within [min, max] inclusive.
> func randInt(min, max int) int {
> 	return rand.Intn(max-min) + min
> ```

Note that the global `rand` functions such as `rand.Intn` internally have a mutex. This actually shows up in profiling of load generators, so `kv` and `ycsb` use a per-goroutine random number generator.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-:-Kqp12Yz84-sLPKSAro2:bnfp4nl)*

rjnn:
:lgtm:, Major kudos for doing this so rapidly! Have you tested this out on a production cluster to see what sort of ops we get?

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[tpcc/ddls.sql, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwxQtGQgBv0dGpMk2b:-KqwxQtGQgBv0dGpMk2c:b-w9hfmz) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/ddls.sql#L1)):*
> ```SQLPL
> create database if not exists tpcc;
> ```

You might want to gate this one on a flag (or just add a TODO), since its unsupported syntax in postgres. We do want to be able to run this load generator against postgres (eventually). 

---

*[tpcc/generate.go, line 121 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwydXjYPcylIkPChOX:-KqwydXjYPcylIkPChOY:b-4wc3rn) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L121)):*
> ```Go
> 	// 100,000 items.
> 	startWorkerPool(func(id int) {
> 		if _, err := stmtItem.Exec(id, randInt(1, 10000), randAString(14, 24), float64(randInt(100, 10000))/float64(100), randOriginalString()); err != nil {
> ```

Could use comments on the arguments the way you've done for the subsequent queries.

---

*[tpcc/generate.go, line 178 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwyoHwEgASUpvzXniP:-KqwyoHxw7Nhot1d2ZG9:bd2ulhc) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L178)):*
> ```Go
> 			// 3000 customers per district
> 			startWorkerPool(func(c_id int) {
> 				// 10% of the customer rows have bad credit.
> ```

This is a big ask to do now (so I don't ask for it), but in the future, when you get nuggets like this off the spec, can you link to the section or page number where the spec says this? When you write the comment you have the information at hand, now I'm searching to figure it out.

---

*[tpcc/main.go, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwzYdszXAN0yrE9K9U:-KqwzYdszXAN0yrE9K9V:b4pkkbj) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/main.go#L67)):*
> ```Go
> 	}
> 
> 	dbURL := "postgresql://root@localhost:26257/tpcc?sslmode=disable"
> ```

Add a flag for `dbURL` since we'll use a proxy in the production clusters. This can be the default.

---

*[tpcc/random.go, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwzkvzI0VXAfRqlv0z:-KqwzkvzI0VXAfRqlv1-:b-cyo3dh) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L28)):*
> ```Go
> 
> var cLastTokens = [...]string{
> 	"BAR", "OUGHT", "ABLE", "PRI", "PRES",
> ```

Is this seriously in the spec :roll_eyes: 

---

*[tpcc/random.go, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwzqBtrhLWJlowc6N2:-KqwzqBtrhLWJlowc6N3:b-3b46dy) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L31)):*
> ```Go
> 	"ESE", "ANTI", "CALLY", "ATION", "EING"}
> 
> // cLoad is the value of C at load time. See 2.1.6.1.
> ```

Just calling out that this comment is exceptional. Makes it very easy to look up the relevant documentation. :ok_hand: :ok_hand: :ok_hand: 

---

*[tpcc/random.go, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqx-2TNO7AgwcIyF0fN:-Kqx-2TNO7AgwcIyF0fO:b-x4o9ds) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L40)):*
> ```Go
> }
> 
> func randFromAlphabet(min, max int, alphabet string) string {
> ```

Better named `randStringFromAlphabet`, and min and max are the lengths, so `minLen` and `maxLen`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-:-Kqx24RXYKmWisdZqei3:b-4wzuc7)*

petermattis:
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[tpcc/main.go, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwzYdszXAN0yrE9K9U:-Kqx603zuadTpZeffz_k:b-147ym9) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/main.go#L67)):*

Add a flag for `dbURL` since we'll use a proxy in the production clusters. This can be the default.
</blockquote></details>

This is configurable via a command line argument, just as it is with `kv` and `ycsb`. See the following 2 lines.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*

rjnn:
*[tpcc/main.go, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwzYdszXAN0yrE9K9U:-Kqx8DGvIjDr1au1CwQ1:bg0tm9k) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/main.go#L67)):*

This is configurable via a command line argument, just as it is with `kv` and `ycsb`. See the following 2 lines.
</blockquote></details>

My bad.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*

rjnn:
*[tpcc/generate.go, line 141 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0YZUEm7VJHYcVsMY:-Kqx9CwW9OsEhFHPY2U9:b-uy0pw6) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L141)):*

Does the spec say anything about how loading occurs? We'd likely see faster loading by batching inserts. If this is kosher, add a `TODO`.
</blockquote></details>

I don't believe it does, from my reading.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*

jordanlewis:
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[tpcc/ddls.sql, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwxQtGQgBv0dGpMk2b:-Kqxu9YbeavyMoV7fzu0:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/ddls.sql#L1)):*

You might want to gate this one on a flag (or just add a TODO), since its unsupported syntax in postgres. We do want to be able to run this load generator against postgres (eventually). 
</blockquote></details>

Done.

---

*[tpcc/generate.go, line 30 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqoy-xHYVjQI4R1qYE1:-KqxuCByf8mlt8QyPMZ6:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L30)):*

`s/sent/set/g`
</blockquote></details>

Done.

---

*[tpcc/generate.go, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp054tp7T9TJkQsVS1:-KqxuD0UXoY8hn8ti3NS:bnbumfo) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L58)):*

With only a single statement, there isn't much reason to be using a transaction vs executing the statement in an implicit transaction.
</blockquote></details>

Can't you still get ambiguous commit errors with implicit transactions? This function protects against those.

---

*[tpcc/generate.go, line 75 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqozFeckv1KlZFulnjP:-Kqxu_QtkI3PWN-lr2Id:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L75)):*

This name implies a function which returns a worker pool (to me, at least). But really what it is does run `loader` `n` times with `*concurrency` workers. Perhaps rename to `parallel` or something like that. Also, small nit: I think calling this function looks better if the function is last. Consider:

```
startWorkerPool(n, "foo", func(int) {
  // load the foos
})
```

vs 

```
startWorkerPool(func(int) {
  // load the foos
}, n "foo")
```
</blockquote></details>

Done.

---

*[tpcc/generate.go, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqoxYPAowB5y9duCKz7:-KqxvLNGoRTb5fjSLDU2:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L99)):*

I think you can use a little bit more formatting in these statements to make them more readable. For example:

```
stmtItem := prepare(db, `
INSERT INTO item (i_id, i_im_id, i_name, i_price, i_data) 
  VALUES ($1, $2, $3, $4, $5)
`)
```
</blockquote></details>

Done.

---

*[tpcc/generate.go, line 121 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwydXjYPcylIkPChOX:-KqxwuDgFEJnrcCU-yTy:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L121)):*

Could use comments on the arguments the way you've done for the subsequent queries.
</blockquote></details>

Done.

---

*[tpcc/generate.go, line 139 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0iDQS-fyLYdcIktt:-Kqxy8xOQm-uZc6A3lyV:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L139)):*

`s/.././g`
</blockquote></details>

Done.

---

*[tpcc/generate.go, line 141 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0YZUEm7VJHYcVsMY:-KqxyDpgnmhFHZoDt14-:b-1k8de4) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L141)):*

I don't believe it does, from my reading.
</blockquote></details>

No, the spec doesn't care about how the data is loaded. The speed of the data loading component is not part of the benchmark.

---

*[tpcc/generate.go, line 178 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwyoHwEgASUpvzXniP:-KqxyeA7fxfRBYzptTvc:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L178)):*

This is a big ask to do now (so I don't ask for it), but in the future, when you get nuggets like this off the spec, can you link to the section or page number where the spec says this? When you write the comment you have the information at hand, now I'm searching to figure it out.
</blockquote></details>

Done.

---

*[tpcc/main.go, line 44 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp-NxF44ec805EKo3Y:-KqxyhpmpjIerMqkUfSd:bhlyrh8) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/main.go#L44)):*

IMO, `log.Printf` is overly verbose due to the prefixed timestamp info. For these load generators, I prefer `fmt.Printf` and including elapsed times when appropriate.
</blockquote></details>

This was cargo culted. Switched to fmt.

---

*[tpcc/random.go, line 28 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqwzkvzI0VXAfRqlv0z:-Kqxyqf8rUjKSqVzk-uk:b2ni4x7) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L28)):*

Is this seriously in the spec :roll_eyes: 
</blockquote></details>

ikr

---

*[tpcc/random.go, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqx-2TNO7AgwcIyF0fN:-Kqxz5w_8zgxH8v36kzH:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L40)):*

Better named `randStringFromAlphabet`, and min and max are the lengths, so `minLen` and `maxLen`.
</blockquote></details>

Done.

---

*[tpcc/random.go, line 94 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqmm0wiPtH9iUBpsISy:-KqxzO-sy6IaI-uNlJlz:by04ygm) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L94)):*

Is this specified? I would have though you'd do `rand.Float64() * 0.2`, though I guess that isn't inclusive at the end range.
</blockquote></details>

The spec says "W_TAX random within [0.0000 .. 0.2000]". Earlier, the spec defines a random range as:

```The term randomly selected within [x .. y] means independently selected at random and uniformly distributed between x and y, inclusively, with a mean of (x+y)/ 2, and with the same number of digits of precision as shown. For example, [0.01 .. 100.00] has 10,000 unique values, whereas [1 ..100] has only 100 unique values.```

This particular implementation is how the Java version I was looking at does it. Perhaps it's incorrect anyway because this really should be a precise decimal, but we're doing floating point arithmetic...
```


---

*[tpcc/random.go, line 99 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqmmW0uqhCoEZiCcrN8:-Kqy-ELgHODMRy79Ihtd:b-70kpcw) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L99)):*

Note that the global `rand` functions such as `rand.Intn` internally have a mutex. This actually shows up in profiling of load generators, so `kv` and `ycsb` use a per-goroutine random number generator.
</blockquote></details>

Noted. Once we get to the actual load generator I'll probably revisit this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*
<!-- Sent from Reviewable.io -->

rjnn:
*[tpcc/random.go, line 94 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqmm0wiPtH9iUBpsISy:-Kqx-nMN9_xwe_bkaiTg:b-c6fc82) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/random.go#L94)):*

The spec says "W_TAX random within [0.0000 .. 0.2000]". Earlier, the spec defines a random range as:

```The term randomly selected within [x .. y] means independently selected at random and uniformly distributed between x and y, inclusively, with a mean of (x+y)/ 2, and with the same number of digits of precision as shown. For example, [0.01 .. 100.00] has 10,000 unique values, whereas [1 ..100] has only 100 unique values.```

This particular implementation is how the Java version I was looking at does it. Perhaps it's incorrect anyway because this really should be a precise decimal, but we're doing floating point arithmetic...
</blockquote></details>

`rand.Float64()` is bad because its not a uniform distribution and should generally be avoided (although due to the truncation to such a small number of significant digits should basically be fine here). I assume that's what the Java version is defensively programming against.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*

petermattis:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[tpcc/generate.go, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp054tp7T9TJkQsVS1:-Kqy30EGUwVXqvvBfZ6i:b-edla84) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L58)):*

Can't you still get ambiguous commit errors with implicit transactions? This function protects against those.
</blockquote></details>

`crdb.ExecuteTx` doesn't protect against ambiguous commit errors. Unless I'm missing something.

---

*[tpcc/generate.go, line 141 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0YZUEm7VJHYcVsMY:-KqyBlvRRt5c0uZWKgZS:bz6a6dx) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L141)):*

No, the spec doesn't care about how the data is loaded. The speed of the data loading component is not part of the benchmark.
</blockquote></details>

Let's add the `TODO`. The speed of loading can affect our dev cycle.

---

*[tpcc/generate.go, line 101 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-KqyBHMI07roCqa58dz0:-KqyBHMI07roCqa58dz1:bkq2uaf) ([raw file](https://github.com/cockroachdb/loadgen/blob/83385ba8acb5a91811f7691551c12f0e0f27109e/tpcc/generate.go#L101)):*
> ```Go
> 	stmtItem := prepare(db, `
> INSERT INTO item (i_id, i_im_id, i_name, i_price, i_data)
> VALUES ($1, $2, $3, $4, $5)`)
> ```

Nit: I think indentation of `VALUES` would help readability, though I suppose below you are placing `VALUES` in column 0 on purpose.  Meh, this isn't terribly important. The formatting is already much better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*

jordanlewis:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[tpcc/generate.go, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp054tp7T9TJkQsVS1:-KqyD_rvHA02CZgbqSgv:b-kbe51q) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L58)):*

`crdb.ExecuteTx` doesn't protect against ambiguous commit errors. Unless I'm missing something.
</blockquote></details>

Err, yeah, you're right. Removed.

---

*[tpcc/generate.go, line 141 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0YZUEm7VJHYcVsMY:-KqyEF7TNR6IxzcEK-Tp:bn2ukpg) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L141)):*

Let's add the `TODO`. The speed of loading can affect our dev cycle.
</blockquote></details>

What kind of batching are you proposing here? Statement-level batching or unrolling into extra clauses in VALUES()? Added the TODO.

BTW, the loading is pretty slow - but not of stock or items. The slowest part is loading the orders, because of all of the FK validation that goes on.

That being said, loading a warehouse now takes about 2-3 minutes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*

petermattis:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[tpcc/generate.go, line 141 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/63#-Kqp0YZUEm7VJHYcVsMY:-KqyF5gWV36PeFvyGIC5:b1nnqqx) ([raw file](https://github.com/cockroachdb/loadgen/blob/a4ffccb323d75ea822b144bdff5d78d4d838b952/tpcc/generate.go#L141)):*

What kind of batching are you proposing here? Statement-level batching or unrolling into extra clauses in VALUES()? Added the TODO.

BTW, the loading is pretty slow - but not of stock or items. The slowest part is loading the orders, because of all of the FK validation that goes on.

That being said, loading a warehouse now takes about 2-3 minutes.
</blockquote></details>

I'm proposing inserting multiple rows per INSERT statement. This should be done for all of the different types of INSERTs as it allows parallelization on the backend.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/63)*

PR #64, opened 2017-08-08
tpcc: add transactions, worker and stats

jordanlewis:
New Order is the backbone transaction of TPCC.
Payments simulates customers making payments.
Order Status is a read-only transaction.

Transactions implemented:
- [x] New order is the backbone transaction of TPCC.
- [x] Payments simulates customers making payments.
- [x] Order status is a read-only transaction that checks a customer's order status.
- [x] Delivery
- [x] Stock level

Also cargo cult some stats and flags from kv. Now we can run a TPCC workload:

```
_elapsed___errors__ops/sec(inst)___ops/sec(cum)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
      1s        0           55.8           55.8    109.1    536.9    637.5    838.9
      2s        0           71.0           63.4    121.6   1006.6   1342.2   1543.5
      3s        0           59.0           61.9    109.1    805.3   1040.2   1040.2
      4s        0           50.1           59.0    151.0   1073.7   1342.2   1342.2
      5s        0           47.0           56.6    159.4   1476.4   3623.9   3623.9
      6s        0           56.0           56.5    130.0   1342.2   1744.8   2013.3
      7s        0           57.9           56.7    142.6    838.9   1543.5   1543.5
      8s        0           59.8           57.1    113.2    973.1    973.1   1006.6
      9s        0           47.2           56.0    121.6    872.4   1275.1   1275.1
     10s        0           57.8           56.2     96.5    671.1   1476.4   2013.3

_elapsed___errors_____ops(total)___ops/sec(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   10.0s        0            562           56.2    258.6    121.6    973.1   1543.5   3623.9

TPCC         562          17803026.1 ns/op
```


jordanlewis:
This seems functional now, but the UI sucks. Any suggestion for improvement?

```
_time______opName__ops/s(inst)__ops/s(cum)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   1s         all        58.9       58.8    176.2    520.1    637.5    704.6
  tpmC (newOrder)        29.9       29.9    121.6    234.9    268.4    268.4
          payment        22.9       22.9    234.9    520.1    637.5    637.5
      orderStatus         1.0        1.0    100.7    100.7    100.7    100.7
         delivery         4.0        4.0    130.0    201.3    201.3    201.3
       stockLevel         1.0        1.0    704.6    704.6    704.6    704.6
   2s         all        55.1       56.9    159.4    872.4   1073.7   1342.2
  tpmC (newOrder)        25.0       27.5    100.7    234.9    268.4    268.4
          payment        23.0       23.0    436.2   1073.7   1342.2   1342.2
      orderStatus         2.0        1.5     15.2     27.3     27.3     27.3
         delivery         4.0        4.0    117.4    184.5    184.5    184.5
       stockLevel         1.0        1.0    604.0    604.0    604.0    604.0
   3s         all        50.0       54.6    151.0   1140.9   1140.9   1140.9
  tpmC (newOrder)        20.0       25.0    113.2    167.8    285.2    285.2
          payment        26.0       24.0    260.0   1140.9   1140.9   1140.9
      orderStatus         2.0        1.7     13.6    100.7    100.7    100.7
         delivery         1.0        3.0    104.9    104.9    104.9    104.9
       stockLevel         1.0        1.0    536.9    536.9    536.9    536.9
   4s         all        54.9       54.7    151.0    838.9   1476.4   1543.5
  tpmC (newOrder)        24.0       24.7    100.7    201.3    251.7    251.7
          payment        23.0       23.7    285.2   1476.4   1543.5   1543.5
      orderStatus         4.0        2.2     62.9    302.0    302.0    302.0
         delivery         1.0        2.5    260.0    260.0    260.0    260.0
       stockLevel         3.0        1.5    637.5    704.6    704.6    704.6
   5s         all        57.1       55.2    142.6    805.3   1879.0   4026.5
  tpmC (newOrder)        25.0       24.8     88.1    159.4    335.5    335.5
          payment        24.0       23.8    285.2   1879.0   4026.5   4026.5
      orderStatus         2.0        2.2     46.1     50.3     50.3     50.3
         delivery         4.0        2.8    109.1    218.1    218.1    218.1
       stockLevel         2.0        1.6    570.4    570.4    570.4    570.4
   6s         all        49.0       54.1    151.0   1275.1   1543.5   1543.5
  tpmC (newOrder)        19.0       23.8    113.2    318.8    419.4    419.4
          payment        26.0       24.2    369.1   1476.4   1543.5   1543.5
      orderStatus         1.0        2.0     28.3     28.3     28.3     28.3
         delivery         2.0        2.7    142.6    142.6    142.6    142.6
       stockLevel         1.0        1.5    520.1    520.1    520.1    520.1
   7s         all        38.9       52.0    260.0    973.1   2818.6   2818.6
  tpmC (newOrder)        17.0       22.8    192.9    419.4    436.2    436.2
          payment        17.0       23.1    419.4   1811.9   2818.6   2818.6
      orderStatus         0.0        1.7      0.0      0.0      0.0      0.0
         delivery         1.0        2.4    159.4    159.4    159.4    159.4
       stockLevel         4.0        1.9    805.3    906.0    906.0    906.0
   8s         all        54.1       52.2    176.2    738.2   1040.2   1879.0
  tpmC (newOrder)        22.0       22.7    113.2    243.3    268.4    268.4
          payment        25.0       23.4    369.1   1040.2   1879.0   1879.0
      orderStatus         3.0        1.9    151.0    218.1    218.1    218.1
         delivery         1.0        2.2    226.5    226.5    226.5    226.5
       stockLevel         3.0        2.0    604.0    738.2    738.2    738.2
   9s         all        44.0       51.3    226.5    838.9   2013.3   2013.3
  tpmC (newOrder)        17.0       22.1    104.9    285.2    302.0    302.0
          payment        22.0       23.2    352.3    939.5   2013.3   2013.3
      orderStatus         2.0        1.9     48.2     79.7     79.7     79.7
         delivery         1.0        2.1    113.2    113.2    113.2    113.2
       stockLevel         2.0        2.0    838.9    838.9    838.9    838.9
  10s         all        52.0       51.4    184.5    872.4   1140.9   3087.0
  tpmC (newOrder)        21.0       22.0    100.7    260.0    369.1    369.1
          payment        22.0       23.1    419.4   1140.9   3087.0   3087.0
      orderStatus         3.0        2.0    134.2    302.0    302.0    302.0
         delivery         3.0        2.2    117.4    335.5    335.5    335.5
       stockLevel         3.0        2.1    771.8    771.8    771.8    771.8
```

petermattis:
> This seems functional now, but the UI sucks. Any suggestion for improvement?

Yeah, don't show the individual op stats during the periodic output unless a flag is specified. You'd still show the individual op stats at exit, regardless of that flag. 

PS Your columns don't quite line up with the headers. 

PPS I'll try to review this tonight or tomorrow.

jordanlewis:
Thanks for the suggestions. One complication is that the actual stat that you're supposed to report in TPCC, `tpmC`, is actually just the throughput of NewOrder transactions. Do you think I should report that one as the primary stat, or still report the throughput of all transactions?

Also, it's not clear whether reporting aggregate transaction latency makes any sense - because each transaction has a completely different latency profile. An aggregate histogram will be really confusing and possibly meaningless.

petermattis:
Ok, I'd report `tmpC` by default then and only display the other ops if some sort of `--verbose` flag is specified. Agreed that per-op latency numbers make more sense.

rjnn:
:lgtm:  Great work! I concur with reporting `tpmC` by default, and displaying the others only when verbose.

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 2 of 3 files at r4, 1 of 2 files at r5, 2 of 2 files at r6, 2 of 2 files at r7, 4 of 4 files at r8, 7 of 7 files at r9.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[tpcc/ddls_postgres.sql, line 1 at r8](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDzAHsjzza0ZredEtj:-KrDzAHsjzza0ZredEtk:b1ro2p0) ([raw file](https://github.com/cockroachdb/loadgen/blob/841b965e837c3430b8051f2ef19c1a5e29b953fa/tpcc/ddls_postgres.sql#L1)):*
> ```SQLPL
> -- Need to run CREATE DATABASE tpcc by hand first.
> ```

I'm less happy about having two completely disjoint DDLs than just having the subset that require different syntax be disjoint.

---

*[tpcc/new_order.go, line 183 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDtolawFWm5aLAbNhq:-KrDtolawFWm5aLAbNhr:b-6bmp6w) ([raw file](https://github.com/cockroachdb/loadgen/blob/ea774e38574869c653735414f87bb98e1a016e64/tpcc/new_order.go#L183)):*
> ```Go
> 
> 		var i_data string
> 		// For each o_ol_cnt item in the order (2.4.2.2)
> ```

This comment could be expanded. Rather than just nitpicking, I want to take this opportunity to instead point out that the _previous_ comments in this function are outstanding, and this one could also be expanded to be so.

---

*[tpcc/new_order.go, line 210 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDu8-BQtzsY-qCDu--:-KrDu8-BQtzsY-qCDu-0:bmd5d67) ([raw file](https://github.com/cockroachdb/loadgen/blob/ea774e38574869c653735414f87bb98e1a016e64/tpcc/new_order.go#L210)):*
> ```Go
> 				d.d_id,
> 				d.w_id,
> 				i+1, // ol_number is a unique number among items that share an ol_o_id
> ```

I don't understand this comment. What does the significance of this being unique mean, given that its incremented?

---

*[tpcc/random.go, line 36 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-Kr2h0tNJIfXTTabot3J:-Kr2h0tNJIfXTTabot3K:b-kjj9tr) ([raw file](https://github.com/cockroachdb/loadgen/blob/7b18a0ef4647c8dd77ac8e7b9df812fc32eddec3/tpcc/random.go#L36)):*
> ```Go
> 
> // cCustomerID is the value of C for the customer id generator. 2.1.6.
> var cCustomerID int
> ```

What is `C`?

---

*[tpcc/worker.go, line 72 at r9](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDzYH5Rrd7vGkldPOw:-KrDzYH5Rrd7vGkldPOx:bb76mu4) ([raw file](https://github.com/cockroachdb/loadgen/blob/6e683d2962572d898a83352a8240697c81f2958d/tpcc/worker.go#L72)):*
> ```Go
> 
> // The weights should add to 100. They match the TPCC spec - so probably don't tune these.
> // Keep this in the same order
> ```

Why?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-:-KrDzcCpBnMojcbeKeQn:b-wh3dwi)*

jordanlewis:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[tpcc/ddls_postgres.sql, line 1 at r8](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDzAHsjzza0ZredEtj:-KrE3RMlZ4enzSA__5ZR:bgnm2gl) ([raw file](https://github.com/cockroachdb/loadgen/blob/841b965e837c3430b8051f2ef19c1a5e29b953fa/tpcc/ddls_postgres.sql#L1)):*

I'm less happy about having two completely disjoint DDLs than just having the subset that require different syntax be disjoint.
</blockquote></details>

Eh, I think we're going to diverge from the standard one so I'd prefer to keep it like this for now. If you care badly we can fix it in a different patch.

---

*[tpcc/new_order.go, line 183 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDtolawFWm5aLAbNhq:-KrE2NjHT92rsb5qvlBj:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/ea774e38574869c653735414f87bb98e1a016e64/tpcc/new_order.go#L183)):*

This comment could be expanded. Rather than just nitpicking, I want to take this opportunity to instead point out that the _previous_ comments in this function are outstanding, and this one could also be expanded to be so.
</blockquote></details>

Done.

---

*[tpcc/new_order.go, line 210 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDu8-BQtzsY-qCDu--:-KrE2O42wyA--u8oVX_A:b-uh6k14) ([raw file](https://github.com/cockroachdb/loadgen/blob/ea774e38574869c653735414f87bb98e1a016e64/tpcc/new_order.go#L210)):*

I don't understand this comment. What does the significance of this being unique mean, given that its incremented?
</blockquote></details>

This is confusing in the spec:

```
A new row is inserted into the ORDER-LINE table to reflect the item on the order. OL_DELIVERY_D is set
to a null value, OL_NUMBER is set to a unique value within all the ORDER-LINE rows that have the same
OL_O_ID value, and OL_DIST_INFO is set to the content of S_DIST_xx, where xx represents the district
number (OL_D_ID)
```

This instruction is scoped to the current transaction, so I think it means that ol_number is just the number of the item in the order - like it's just saying that every order line row with a given order id has to have a different ol_number.

Anyway I updated the comment to be more clear.

---

*[tpcc/random.go, line 36 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-Kr2h0tNJIfXTTabot3J:-KrE2sJEbKeuYHAk_Ny8:bglsvb5) ([raw file](https://github.com/cockroachdb/loadgen/blob/7b18a0ef4647c8dd77ac8e7b9df812fc32eddec3/tpcc/random.go#L36)):*

What is `C`?
</blockquote></details>

This is a parameter for the non-uniform random generator that's supposed to be varied during the course of a load run in such a way that "does not affect performance". I don't really understand the significance of it, and it looks like other generators don't actually vary C throughout the course of load either.

From the spec:

```
2.1.6 The term non-uniform random, used only for generating customer numbers, customer last names,
and item numbers, means an independently selected and non -uniformly distributed random number over the
specified range of values [x .. y]. This number must be generated by using the function NURand which produces
positions within the range [x .. y]. The results of NURand might have to be converted to produce a name or a
number valid for the implementation.
NURand(A, x, y) = (((random(0, A) | random(x, y)) + C) % (y - x + 1)) + x
where:
exp-1 | exp-2 stands for the bitwise logical OR operation between exp -1 and exp-2
exp-1 % exp-2 stands for exp-1 modulo exp-2
random(x, y) stands for randomly selected within [x .. y]
A is a constant chosen according to the size of the range [x .. y]
for C_LAST, the range is [0 .. 999] and A = 255
for C_ID, the range is [1 .. 3000] and A = 1023
for OL_I_ID, the range is [1 .. 100000] and A = 8191
C is a run-time constant randomly chosen within [0 .. A] that can be varied without altering performance.
The same C value, per field (C_LAST, C_ID, and OL_I_ID), must be used by all emulated terminals.
2.1.6.1 In order that the value of C used for C_LAST does not alter performance the following must be true:
Let C-Load be the value of C used to generate C_LAST when populating the database. C-Load is a value
in the range of [0..255] including 0 and 255.
TPC Benchmark™ C - Standard Specification, Revision 5.11 - Page 22 of 130
Let C-Run be the value of C used to generate C_LAST for the measurement run.
Let C-Delta be the absolute value of the difference between C-Load and C-Run. C-Delta must be a value in
the range of [65..119] including the values of 65 and 119 and excluding the value of 96 and 112.
```

---

*[tpcc/worker.go, line 72 at r9](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDzYH5Rrd7vGkldPOw:-KrE3OKEdyc-0rzlUoWo:b21oghr) ([raw file](https://github.com/cockroachdb/loadgen/blob/6e683d2962572d898a83352a8240697c81f2958d/tpcc/worker.go#L72)):*

Why?
</blockquote></details>

Got rid of the order clause which actually doesn't matter anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/64)*

jordanlewis:
Okay, done - tpmC is now the only stat output.

```
_time______opName__ops/s(inst)__ops/s(cum)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   1s        tpmC         18.9        18.9    113.2    218.1    335.5    335.5
   2s        tpmC         30.0        24.4    176.2    335.5    436.2    436.2
   3s        tpmC         21.0        23.3    209.7    352.3    402.7    402.7
   4s        tpmC         24.0        23.5    125.8    260.0    469.8    469.8
   5s        tpmC         26.0        24.0    134.2    243.3    268.4    268.4
   6s        tpmC         24.0        24.0    151.0    469.8    486.5    486.5
   7s        tpmC         20.0        23.4    159.4   1006.6   1040.2   1040.2
   8s        tpmC         24.0        23.5    226.5    369.1    419.4    419.4
   9s        tpmC         23.1        23.4    125.8    285.2    302.0    302.0
^C
_elapsed___newOrders___tpmC(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
    9.2s         215        23.3    178.6    151.0    402.7    486.5   1040.2

TPCC         215          42907303.5 ns/op
```

---

Review status: 8 of 11 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-:-KrE6_Bt53aTqFsujrkv:bbisas6)*

rjnn:
Reviewed 3 of 3 files at r10.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[tpcc/ddls_postgres.sql, line 1 at r8](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDzAHsjzza0ZredEtj:-KrEA9gvFmotf6FiN1Vr:bkhjo7o) ([raw file](https://github.com/cockroachdb/loadgen/blob/841b965e837c3430b8051f2ef19c1a5e29b953fa/tpcc/ddls_postgres.sql#L1)):*

Eh, I think we're going to diverge from the standard one so I'd prefer to keep it like this for now. If you care badly we can fix it in a different patch.
</blockquote></details>

No I don't insist, I trust your call on this.

---

*[tpcc/new_order.go, line 183 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDtolawFWm5aLAbNhq:-KrEAE2RTSbf9LyysRIa:b-1vahzw) ([raw file](https://github.com/cockroachdb/loadgen/blob/ea774e38574869c653735414f87bb98e1a016e64/tpcc/new_order.go#L183)):*

Done.
</blockquote></details>

Wonderful!

---

*[tpcc/new_order.go, line 210 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrDu8-BQtzsY-qCDu--:-KrEAMB9YCfeE2Ht-uHd:b3bh1xw) ([raw file](https://github.com/cockroachdb/loadgen/blob/ea774e38574869c653735414f87bb98e1a016e64/tpcc/new_order.go#L210)):*

This is confusing in the spec:

```
A new row is inserted into the ORDER-LINE table to reflect the item on the order. OL_DELIVERY_D is set
to a null value, OL_NUMBER is set to a unique value within all the ORDER-LINE rows that have the same
OL_O_ID value, and OL_DIST_INFO is set to the content of S_DIST_xx, where xx represents the district
number (OL_D_ID)
```

This instruction is scoped to the current transaction, so I think it means that ol_number is just the number of the item in the order - like it's just saying that every order line row with a given order id has to have a different ol_number.

Anyway I updated the comment to be more clear.
</blockquote></details>

Excellent. This is far clearer.

---

*[tpcc/random.go, line 36 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-Kr2h0tNJIfXTTabot3J:-KrEAUaHyQAQ2CLneALL:bgvt1ym) ([raw file](https://github.com/cockroachdb/loadgen/blob/7b18a0ef4647c8dd77ac8e7b9df812fc32eddec3/tpcc/random.go#L36)):*

This is a parameter for the non-uniform random generator that's supposed to be varied during the course of a load run in such a way that "does not affect performance". I don't really understand the significance of it, and it looks like other generators don't actually vary C throughout the course of load either.

From the spec:

```
2.1.6 The term non-uniform random, used only for generating customer numbers, customer last names,
and item numbers, means an independently selected and non -uniformly distributed random number over the
specified range of values [x .. y]. This number must be generated by using the function NURand which produces
positions within the range [x .. y]. The results of NURand might have to be converted to produce a name or a
number valid for the implementation.
NURand(A, x, y) = (((random(0, A) | random(x, y)) + C) % (y - x + 1)) + x
where:
exp-1 | exp-2 stands for the bitwise logical OR operation between exp -1 and exp-2
exp-1 % exp-2 stands for exp-1 modulo exp-2
random(x, y) stands for randomly selected within [x .. y]
A is a constant chosen according to the size of the range [x .. y]
for C_LAST, the range is [0 .. 999] and A = 255
for C_ID, the range is [1 .. 3000] and A = 1023
for OL_I_ID, the range is [1 .. 100000] and A = 8191
C is a run-time constant randomly chosen within [0 .. A] that can be varied without altering performance.
The same C value, per field (C_LAST, C_ID, and OL_I_ID), must be used by all emulated terminals.
2.1.6.1 In order that the value of C used for C_LAST does not alter performance the following must be true:
Let C-Load be the value of C used to generate C_LAST when populating the database. C-Load is a value
in the range of [0..255] including 0 and 255.
TPC Benchmark™ C - Standard Specification, Revision 5.11 - Page 22 of 130
Let C-Run be the value of C used to generate C_LAST for the measurement run.
Let C-Delta be the absolute value of the difference between C-Load and C-Run. C-Delta must be a value in
the range of [65..119] including the values of 65 and 119 and excluding the value of 96 and 112.
```
</blockquote></details>

It sounds to me like it's an 'anti-cheating' mechanism. Thanks for clarifying!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/64)*

petermattis:
:lgtm:

This is great! Well organized and commented. A few minor comments below. I didn't take a look at all of the op implementations in detail. That can wait for when we start analyzing where our performance is going. 

---

Review status: all files reviewed at latest revision, 10 unresolved discussions, all commit checks successful.

---

*[tpcc/ddls.sql, line 1 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGJgkOgzKbwZavYL5k:-KrGJgkOgzKbwZavYL5l:b-bihpb2) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/ddls.sql#L1)):*
> ```SQLPL
> create database if not exists tpcc;
> ```

Should this file be named `ddls_cockroach.sql`?

---

*[tpcc/delivery.go, line 55 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGK0ohyNQ0WBepagGb:-KrGK0ohyNQ0WBepagGc:b-7j7iwd) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L55)):*
> ```Go
> 			WHERE no_w_id = $1 AND no_d_id = $2
> 			ORDER BY no_o_id ASC
> 			LIMIT 1`)
> ```

Each of these prepare statements requires a round-trip to the server. Better to prepare the statement on the `db` and then call `tx.Stmt`, though I believe there are inefficiencies with that in go1.8. I'd leave this as-is for now, just pointing it out as something we can optimize.

---

*[tpcc/delivery.go, line 61 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGKnlo8yGOdiFlLUeB:-KrGKnlo8yGOdiFlLUeC:b-ydetrj) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L61)):*
> ```Go
> 		delNewOrder, err := tx.Prepare(`
> 			DELETE FROM new_order
> 			WHERE no_w_id = $1 AND no_d_id = $2 AND no_o_id = $3`)
> ```

Can you get rid of `getNewOrder` by doing something like:

```
  DELETE FROM new_order WHERE ... LIMIT 1 RETURNING no_o_id
```

---

*[tpcc/delivery.go, line 89 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGLSu3vAXIzeqGnf5q:-KrGLSu3vAXIzeqGnf5r:b-13s9di) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L89)):*
> ```Go
> 			UPDATE customer
> 			SET (C_BALANCE, C_DELIVERY_CNT) =
> 				(C_BALANCE + $1, C_DELIVERY_CNT + 1)
> ```

Lowercase `C_BALANCE` and `C_DELIVERY_CNT` for consistency.

---

*[tpcc/delivery.go, line 100 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGKc9Oy_T-pNiqy-tp:-KrGKc9Oy_T-pNiqy-tq:b-iyzkx8) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L100)):*
> ```Go
> 			if err := getNewOrder.QueryRow(w_id, d_id).Scan(&o_id); err != nil {
> 				// If no matching order is found, the delivery of this order is skipped.
> 				// TODO(jordan): make sure the error is a not-found
> ```

Check for `err == sql.ErrNoRows`.

---

*[tpcc/main.go, line 37 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGLfIqHwgaT5_AsFDf:-KrGLfIqHwgaT5_AsFDg:b-bkhd57) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/main.go#L37)):*
> ```Go
> 
> var concurrency = flag.Int("concurrency", 2*runtime.NumCPU(), "Number of concurrent writers inserting blocks")
> var ddls = flag.String("ddls", "ddls.sql", "which ddls file to load")
> ```

We probably want to embed the ddls in the binary. Otherwise we have to ship them with the load generator which isn't something we do with other load generators. Take a look at `tpch` for how the queries are embedded. Seems relatively straightforward to have `cockroachDDLs` and `postgresDDLs` globals.

---

*[tpcc/main.go, line 48 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGLy2VKNnnBhSUfU3g:-KrGLy2VKNnnBhSUfU3h:bvl0pr2) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/main.go#L48)):*
> ```Go
> 
> // outputInterval = interval at which information is output to console.
> var outputInterval = flag.Duration("output-interval", 1*time.Second, "Interval of output")
> ```

I'm sure this was cargo-culted from somewhere else, but this is a flag that is never changed. Let's just hardcode 1s. Do you mind removing it from wherever you took it as well?

---

*[tpcc/main.go, line 52 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGM59-gxIWGd-0W7tW:-KrGM59-gxIWGd-0W7tX:bw16mrt) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/main.go#L52)):*
> ```Go
> const (
> 	minLatency = 100 * time.Microsecond
> 	maxLatency = 10 * time.Second
> ```

It is possible this `maxLatency` is too small for tpcc ops. Just something to keep in mind if we start seeing 10s+ latencies.

---

*[tpcc/random.go, line 38 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGMe_Orc62_JbzNgDl:-KrGMe_Orc62_JbzNgDm:b7rj88g) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/random.go#L38)):*
> ```Go
> var cCustomerID int
> 
> //
> ```

Spurious blank comment line.

---

*[tpcc/stock_level.go, line 72 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGMjjvgVdmwg-cpiwf:-KrGMjjvgVdmwg-cpiwg:b5ig240) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/stock_level.go#L72)):*
> ```Go
> 		// the threshold.
> 		if err := tx.QueryRow(`
> 				SELECT COUNT( DISTINCT (s_i_id))
> ```

`s/COUNT( DISTINCT/COUNT(DISTINCT/g`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-:-KrGN4565bMwfduvuFPf:bb35keh)*

jordanlewis:
TFTRs

---

Review status: 3 of 10 files reviewed at latest revision, 10 unresolved discussions.

---

*[tpcc/ddls.sql, line 1 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGJgkOgzKbwZavYL5k:-KrGvf3WhUEHtTp0A1Ej:b-k7mx7q) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/ddls.sql#L1)):*

Should this file be named `ddls_cockroach.sql`?
</blockquote></details>

Deleted.

---

*[tpcc/delivery.go, line 55 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGK0ohyNQ0WBepagGb:-KrGq5QfSWokgKEBa0Uv:b-pebzh8) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L55)):*

Each of these prepare statements requires a round-trip to the server. Better to prepare the statement on the `db` and then call `tx.Stmt`, though I believe there are inefficiencies with that in go1.8. I'd leave this as-is for now, just pointing it out as something we can optimize.
</blockquote></details>

Yeah, I wanted to do that but I read something that said that you shouldn't prepare on the `db` and use within a transaction because of those "inefficiencies". I'm not sure whether those inefficiencies are like extra copies or something worse like not actually reusing the server side prepared statement. Agreed that we should look into this later.

---

*[tpcc/delivery.go, line 61 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGKnlo8yGOdiFlLUeB:-KrGpPynMWi9BkryaGux:b-n8fy6f) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L61)):*

Can you get rid of `getNewOrder` by doing something like:

```
  DELETE FROM new_order WHERE ... LIMIT 1 RETURNING no_o_id
```
</blockquote></details>

DELETE FROM doesn't permit an ORDER BY, which is why I didn't do it that way. I suppose it should always return the smallest one even without an ORDER BY, but it feels a little sketchy.

---

*[tpcc/delivery.go, line 89 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGLSu3vAXIzeqGnf5q:-KrGnS_Kv2a85WWj8TN4:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L89)):*

Lowercase `C_BALANCE` and `C_DELIVERY_CNT` for consistency.
</blockquote></details>

Done.

---

*[tpcc/delivery.go, line 100 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGKc9Oy_T-pNiqy-tp:-KrGnbFyfWkbDGKWt30y:b9rrkvj) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L100)):*

Check for `err == sql.ErrNoRows`.
</blockquote></details>

Thanks, saved me a trip to the docs:)

---

*[tpcc/main.go, line 37 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGLfIqHwgaT5_AsFDf:-KrGo6dsxpIpz18PONhy:bk09s6y) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/main.go#L37)):*

We probably want to embed the ddls in the binary. Otherwise we have to ship them with the load generator which isn't something we do with other load generators. Take a look at `tpch` for how the queries are embedded. Seems relatively straightforward to have `cockroachDDLs` and `postgresDDLs` globals.
</blockquote></details>

Yeah, okay. Done.

---

*[tpcc/main.go, line 48 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGLy2VKNnnBhSUfU3g:-KrGnsDnihpw6kb5ieBC:b-k7mx7q) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/main.go#L48)):*

I'm sure this was cargo-culted from somewhere else, but this is a flag that is never changed. Let's just hardcode 1s. Do you mind removing it from wherever you took it as well?
</blockquote></details>

Deleted.

---

*[tpcc/main.go, line 52 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGM59-gxIWGd-0W7tW:-KrGnut7C5voTUpz0nn3:b-b1cmpk) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/main.go#L52)):*

It is possible this `maxLatency` is too small for tpcc ops. Just something to keep in mind if we start seeing 10s+ latencies.
</blockquote></details>

Yeah, noted.

---

*[tpcc/random.go, line 38 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGMe_Orc62_JbzNgDl:-KrGo3DPwne3S0QhnWI2:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/random.go#L38)):*

Spurious blank comment line.
</blockquote></details>

Done.

---

*[tpcc/stock_level.go, line 72 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGMjjvgVdmwg-cpiwf:-KrGo0klwoa8CUz5fr6Q:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/stock_level.go#L72)):*

`s/COUNT( DISTINCT/COUNT(DISTINCT/g`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-:-KrGvcudhxW06o2gluva:b-814g6a)*

petermattis:
Review status: 3 of 11 files reviewed at latest revision, 10 unresolved discussions.

---

*[tpcc/delivery.go, line 55 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGK0ohyNQ0WBepagGb:-KrH-cCVYTkixxoKwwa3:b-ewj05h) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L55)):*

Yeah, I wanted to do that but I read something that said that you shouldn't prepare on the `db` and use within a transaction because of those "inefficiencies". I'm not sure whether those inefficiencies are like extra copies or something worse like not actually reusing the server side prepared statement. Agreed that we should look into this later.
</blockquote></details>

Another thing to investigate is the `?binary_parameters` `lib/pq` driver option. Using that, we wouldn't prepare the statement but simply execute it with parameters.

---

*[tpcc/delivery.go, line 61 at r10](https://reviewable.io:443/reviews/cockroachdb/loadgen/64#-KrGKnlo8yGOdiFlLUeB:-KrH-jMNeGhI_E5nDqxY:b809xsg) ([raw file](https://github.com/cockroachdb/loadgen/blob/62ba74957d6bcad905e1daea09f12fc5e2e9afc1/tpcc/delivery.go#L61)):*

DELETE FROM doesn't permit an ORDER BY, which is why I didn't do it that way. I suppose it should always return the smallest one even without an ORDER BY, but it feels a little sketchy.
</blockquote></details>

You could use a subquery:

```
DELETE FROM new_order WHERE 
  no_w_id = $1 AND no_d_id = $2 AND no_o_od = (SELECT no_o_id FROM new_order ... ORDER BY no_o_id LIMIT 1) 
  RETURNING no_o_id
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/64)*

PR #66, opened 2017-08-15
kv: divide-and-conquer approach to splitting

petermattis:
Scatter the replicas after each split to get a balanced state quicker.

### DIFF
--- a/kv/main.go
+++ b/kv/main.go
@@ -30,6 +30,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"syscall"
@@ -297,6 +298,12 @@ func (c *cockroach) clone() database {
 	return c
 }
 
+type int64Slice []int64
+
+func (p int64Slice) Len() int           { return len(p) }
+func (p int64Slice) Less(i, j int) bool { return p[i] < p[j] }
+func (p int64Slice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
 func setupCockroach(parsedURL *url.URL) (database, error) {
 	// Open connection to server and create a database.
 	db, dbErr := sql.Open("postgres", parsedURL.String())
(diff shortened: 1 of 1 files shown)


petermattis:
Should probably use this same approach in `ycsb`. Or this might be a good candidate for pulling out into a shared package.

kv/main.go
```diff
@@ -328,15 +329,64 @@ func setupCockroach(parsedURL *url.URL) (database, error) {
...
+					if !ok {
+						break
+					}
+					m := (p.lo + p.hi) / 2
+					split := splitPoints[m]
+					if _, err := db.Exec(`ALTER TABLE test.kv SPLIT AT VALUES ($1)`, split); err != nil {
+						log.Fatal(err)
+					}
+					if _, err := db.Exec(fmt.Sprintf(`ALTER TABLE test.kv SCATTER FROM (%d) TO (%d)`,
+						splitPoints[p.lo], split+1)); err != nil {
```
a-robinson: Minor, but a comment here would make this clearer. I assume the `+1` on `split` is to make it hit both sides of the split, but it took some thinking to understand the `+1`.

a-robinson (approved):
LGTM

kv/main.go
petermattis: Heh, it took some thinking to get that code right too. Done.

kv/main.go
```diff
@@ -328,15 +329,66 @@ func setupCockroach(parsedURL *url.URL) (database, error) {
...
+						}
+					}()
+				}
+			}()
 		}
+
+		for i := 0; i < len(splitPoints); i++ {
+			<-doneCh
+			if (i+1)%1000 == 0 {
+				fmt.Printf("%d splits\n", i+1)
```
rjnn, line 391: Could stick a `printf` after the loop as well, given that someone may use a split number thats not a multiple of 1000.

rjnn (approved):
LGTM.

kv/main.go
petermattis: The purpose is to be a progress indicator. We'll start running the test at that point giving adequate indication of progress.

PR #68, opened 2017-08-16
build: don't need deps during push

jordanlewis:
Also improve tpcc .gitignore


rjnn (approved):
LGTM, but 🔬 🐶  on vendoring. I checked out this PR locally, and `make` did all the right things, so apparently vendoring worked.

PR #69, opened 2017-08-16
tpcc: set isolation level to serializable in txns

rjnn:
This commit adds support for explicitly setting the isolation level. Doing this work revealed that Cockroach doesn't actually support this syntax (although this syntax really is to ensure serializability in Postgres so that we are comparing apples to apples), so this isn't ready for merging until that is fixed.

### DIFF
--- a/tpcc/delivery.go
+++ b/tpcc/delivery.go
@@ -51,7 +51,7 @@ func (del delivery) run(db *sql.DB, wID int) (interface{}, error) {
 	if err := crdb.ExecuteTx(
 		context.Background(),
 		db,
-		&sql.TxOptions{},
+		&sql.TxOptions{Isolation: sql.LevelSerializable},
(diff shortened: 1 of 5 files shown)


andreimatei:
which syntax don't we support? `SET TRANSACTION ISOLATION LEVEL` is supported.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/69#-:-Krh94liRAYFAkRuUefh:buxdlnl)*

jordanlewis:
I believe neither the READ nor WRITE keywords are supported there.

rjnn:
https://github.com/cockroachdb/cockroach/issues/17709

jordanlewis (approved):

PR #72, opened 2017-08-17
dep: update dependencies

rjnn:


tbg:
Rationale? Also seems too late to bump deps across the board.

tbg:
Oh wait, this is `loadgen`! Disregard me. :)

rjnn:
Rationale: we need the new sig in `crdb.ExecuteTx` which takes a `*sql.TxOptions` so that we can pass in the appropriate isolation level (i.e., we want to make sure we are running our transactions at a serializable isolation level so that our comparisons to Postgres are apples to apples, and we can bump that across the board, now that cockroachdb/cockroach#17709 is merged and this syntax is supported in CockroachDB)

jordanlewis:
LGTM

PR #73, opened 2017-08-17
build: mount bin as a tempdir

jordanlewis:
To prevent agent pollution.


tbg:
Wow, this 🐰 🕳 is deep!

tbg (approved):
... but I don't have a lot of context.

benesch (approved):

teamcity-test.sh
```diff
@@ -6,5 +6,6 @@ docker run \
     --workdir=/go/src/github.com/cockroachdb/loadgen \
     --volume="${GOPATH%%:*}/src":/go/src \
     --volume="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)":/go/src/github.com/cockroachdb/loadgen \
```
benesch, line 8: can you tack on `:ro` here so errant writes fail the build instead of bricking the agent?

Issue #48, opened 2017-04-24
tpch: add primary keys to the ddls.go


rjnn:
The TPC-H DDLs currently do not specify primary key columns for the table, but the TPC-H specification (see page 14) has an explicit requirement for what the primary key must be for each table.

rjnn:
This was done when TPC-H was moved to the main cockroach repository.

PR #77, opened 2017-08-22
add explicit primary keys to tpch schema

dt:


rjnn:
So I didn't do this because the spec only says that we need to enforce the unique index constraints, so I did the minimal thing. I also postponed adding this since it creates a discontinuity in our comparisons over time. Can you do a before/after test of the TPC-H queries and log the running times?

Also, up the commit message to "closes #48".

:LGTM:

---

Reviewed 1 of 1 files at r1.

---

*[tpch/ddls.go, line 127 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/77#-Ks9YNobIr6jBXAbyJfi:-Ks9YNocSvQ5fxuBGVRm:b-lkoupt) ([raw file](https://github.com/cockroachdb/loadgen/blob/619e5b50615a04af461e77f9176623c68fb8d13e/tpch/ddls.go#L127)):*
> ```Go
> 	lineitem: `
>     CREATE TABLE lineitem (
>       l_orderkey      INTEGER NOT NULL,
> ```

`PRIMARY KEY (l_orderkey, l_linenumber)` is also in the Spec section 1.4.2.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/77#-:-Ks9Y_PjHgNHkjzeww95:b6biimq)*

dt:
Since the spec allows those indexes (as long as they store only fields from
their tables), it seems like we're in the clear using them, and as far as
minimal, I'd argue that *this* is minimal, as it avoids adding and extra
col and indirect index. I haven't benchmarked (was hoping to get your help
on that) but I suspect it's faster.

dt:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[tpch/ddls.go, line 127 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/77#-Ks9YNobIr6jBXAbyJfi:-KsA0AARA1qrEL18ZxKP:bgsfe04) ([raw file](https://github.com/cockroachdb/loadgen/blob/619e5b50615a04af461e77f9176623c68fb8d13e/tpch/ddls.go#L127)):*

`PRIMARY KEY (l_orderkey, l_linenumber)` is also in the Spec section 1.4.2.
</blockquote></details>

looks like we didn't have that one in our unique indexes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/77)*

rjnn:
Yup, this is better. I just didn't know better when I wrote the load generator.

Just do a scalefactor=1 test run on your laptop, that should be sufficient.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[tpch/ddls.go, line 127 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/77#-Ks9YNobIr6jBXAbyJfi:-KsA6v_EeC-ZxU_sy3yi:b1d521s) ([raw file](https://github.com/cockroachdb/loadgen/blob/619e5b50615a04af461e77f9176623c68fb8d13e/tpch/ddls.go#L127)):*
<details><summary><i>Previously, dt (David Taylor) wrote…</i></summary><blockquote>

looks like we didn't have that one in our unique indexes?
</blockquote></details>

Yes, that was an omission on my part, but now that you're fixing things... :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/77#-:-KsA6ruVXBV-MKy6Nq7H:b-2liymw)*

rjnn:
Ping.

PR #83, opened 2017-09-30
Implement --batch for cockroach read operations

petermattis:
Previously, --batch was silently ignored for read operations and we'd
always read a single key even though workers were incrementing the
number of ops performed as if multiple keys has been read.

For a 95% read workload, using a larger batch size can significantly
improve throughput while bumping up concurrency by the same amount has a
much smaller effect. The numbers below are all taken against a single
node cluster on my laptop:

```
--read-percent=95 --splits=100 --batch=1 --concurrency=16
_elapsed___errors_____ops(total)___ops/sec(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   60.0s        0        1205235        20086.5      0.8      0.6      2.2      4.1     31.5

--read-percent=95 --splits=100 --batch=1 --concurrency=160
_elapsed___errors_____ops(total)___ops/sec(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   60.0s        0        1153993        19230.3      8.3      1.6     30.4     62.9    142.6

--read-percent=95 --splits=100 --batch=10 --concurrency=16
_elapsed___errors_____ops(total)___ops/sec(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   60.0s        0        2631580        43856.6      3.6      2.4     10.0     28.3     79.7

--read-percent=95 --splits=100 --batch=10 --concurrency=4
_elapsed___errors_____ops(total)___ops/sec(cum)__avg(ms)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)
   60.0s        0        2172310        36205.0      1.1      1.0      2.1      5.0     83.9
```

### DIFF
--- a/kv/main.go
+++ b/kv/main.go
@@ -204,7 +204,7 @@ func randomBlock(r *rand.Rand) []byte {
 }
 
 type database interface {
-	read(key int64) error
+	read(count int, g generator) error
 	write(count int, g generator) error
 	clone() database
 }
@@ -244,7 +244,7 @@ func (b *blocker) run(errCh chan<- error, wg *sync.WaitGroup, limiter *rate.Limi
 		start := time.Now()
 		var err error
 		if b.gen.rand().Intn(100) < *readPercent {
-			err = b.db.read(b.gen.readKey())
+			err = b.db.read(*batch, b.gen)
 		} else {
 			err = b.db.write(*batch, b.gen)
 		}
@@ -271,9 +271,18 @@ type cockroach struct {
 	writeStmt *sql.Stmt
 }
 
-func (c *cockroach) read(k int64) error {
-	var v []byte
-	if err := c.readStmt.QueryRow(k).Scan(&k, &v); err != nil && err != sql.ErrNoRows {
+func (c *cockroach) read(count int, g generator) error {
+	args := make([]interface{}, count)
+	for i := 0; i < count; i++ {
+		args[i] = g.readKey()
+	}
+	rows, err := c.readStmt.Query(args...)
+	if err != nil {
+		return err
+	}
+	for rows.Next() {
+	}
+	if err := rows.Err(); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


petermattis:
@spencerkimball this meshes with your observation the other day. I'm not sure how we can take advantage of this observation (yet), but it highlights that parts of our system are able to read rows much faster than we're currently seeing. 

Cc @a-robinson, @nvanbenschoten, @jordanlewis

tbg (approved):

rjnn:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/83#-:-KvI5aAGbGrT09n10Hea:bnfp4nl)*

jordanlewis:
:lgtm:

Have we been testing and recording results with batch!=1? If so, could you summarize how you expect this source of measurement error might have affected those results?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/83#-:-KvI9KKI9KCrkqGm1D--:b-7hesuc)*

petermattis:
> Have we been testing and recording results with batch!=1? If so, could you summarize how you expect this source of measurement error might have affected those results?

No, we haven't. In the past, I've tested with `batch != 1` for writes, but never for reads. 

PS What is going on with teamcity?

jordanlewis:
Not sure why TC didn't pick up the build. You can ask it to check for pending changes if that happens again. It shouldn't be necessary in general though. I don't think I've socialized that yet - I'll add it to the wiki.
image

spencerkimball:
@petermattis if you up the batch size to 100, you can push about 4x the non-batched performance for 99% read load. That seems to be the upper limit if we can find a way to batch non-txnal reads from workloads with high locality.

petermattis:
> That seems to be the upper limit if we can find a way to batch non-txnal reads from workloads with high locality.

Batching the reads is one approach. There might be others. I'm interested in understanding the bottleneck better as this PR highlights that something is problematic in the area encompassing sql down to dist-sender. Note that because of the splits and the random reads, we're sending the same number of RPCs.

petermattis:
Also, need to validate that this batching is helpful on an actual cluster.

PR #89, opened 2017-10-24
kv: add support for multiple nodes

petermattis:
Round-robin amongst the specified nodes.


petermattis:
This will allow testing against different cluster sizes without going through haproxy setup. haproxy isn't onerous, but I'd rather avoid the necessary scripting when running a bunch of performance tests.

mberhault (approved):

PR #95, opened 2017-10-30
rand: add support for "INT8", "BIGINT", "INT", "INTEGER", "INT4", "SM…

robert-s-lee:
"INT8", "BIGINT" 64bits 
"INT", "INTEGER", "INT4" 32bits 
"SMALLINT", "INT2" 16bits

### DIFF
--- a/rand/main.go
+++ b/rand/main.go
@@ -152,6 +152,11 @@ func (b *worker) run(errCh chan<- error, wg *sync.WaitGroup, limiter *rate.Limit
 					case "BIGINT", "INT8":
 						params[k] = rand.Int63()
 					case "INT", "INTEGER", "INT4":
+						// using Postgres specs on integer range so that test data could be used for Postgres and CockroachDB
+						// https://www.postgresql.org/docs/current/static/datatype-numeric.html
+						// https://www.cockroachlabs.com/docs/stable/int.html
+						// Type         CRDB   Postgres
+						// INT, Integer 64bits 32bits
 						params[k] = rand.Int31()
 					case "SMALLINT", "INT2":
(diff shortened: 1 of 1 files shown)


bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[rand/main.go, line 154 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/95#-KxjFR8eJ6bLRfi4h0jj:-KxjFR8eJ6bLRfi4h0jk:b-jccprd) ([raw file](https://github.com/cockroachdb/loadgen/blob/f952400a158c639a4c750d94eb61573049bcf712/rand/main.go#L154)):*
> ```Go
> 					case "BIGINT", "INT8":
> 						params[k] = rand.Int63()
> 					case "INT", "INTEGER", "INT4":
> ```

INT and INTEGER are 64 bits. Only INT4 is 32 bits. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/95)*

robert-s-lee:
Using Postgres specs on data range so that data could be used for Postgres and CockroachDB

bdarnell:
:lgtm:

Add a comment that that's why you're not using the full range of the cockroach `int` type.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/95#-:-KxjJMTW4hNdNGjG5Xe8:btic9bp)*

jordanlewis (approved):

PR #102, opened 2017-11-14
ycsb: pull various improvements from kv

petermattis:
Allow specifying multiple databases to talk to on the command line.

Copy the kv splitting code which splits and scatters in parallel.

### DIFF
--- a/ycsb/main.go
+++ b/ycsb/main.go
@@ -34,6 +34,8 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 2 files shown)


petermattis:
It might be time soon to investigate sharing code between `ycsb`, `kv` and other load generators. But not yet.

ycsb/main.go
```diff
@@ -501,8 +554,10 @@ func (m *mongo) clone() database {
 	}
 }
 
-func setupMongo(parsedURL *url.URL) (database, error) {
-	session, err := mgo.Dial(parsedURL.String())
+func setupMongo(dbURLs []string) (database, error) {
+	// NB: Mongo automatically detects the other nodes in the cluster. We just
```
rjnn: s/Mongo/the Mongo driver/

rjnn (approved):
LGTM

ycsb/main.go
petermattis: Done.

PR #103, opened 2017-11-14
ycsb: don't override cluster read consistency setting on reads

petermattis:
We were overriding the cluster read consistency setting for reads and
always using a read consistency of ONE. Doing so was making the
--cassandra-consistency flag a lie.


petermattis:
`git blame` says this was my fault. The very earliest support for cassandra had this code.

jordanlewis (approved):

a-robinson (approved):

PR #106, opened 2017-11-29
added benchmark for interleaved tables

richardwu:


interleavebench/main.go
```diff
@@ -0,0 +1,324 @@
...
+		}
+
+	}
+}
+
+var stderr = log.New(os.Stderr, "", 0)
+
+func defaultURL(name string) string {
+	if *dbms == "postgres" {
+		return fmt.Sprintf("postgres://richardwu@localhost:5432/%s?sslmode=disable", name)
```
petermattis: Do you need to include your username here?

petermattis: Also, the usual practice for load generators it to take a URL on the command line rather than a flag. This allows you to point the generator at a remote node.

interleavebench/ddls.go
```diff
@@ -0,0 +1,235 @@
...
+var interleaveName = map[tableName]tableName{
+	Product: ProductInterleaved,
+	Variant: VariantInterleaved,
+	Store:   StoreInterleaved,
+}
+
+// Map of all DDL statements that are run when loadSchema is invoked.
+var ddlStmts = map[tableName]string{
+	Merchant: `
+	create table merchant (
```
petermattis: Nit: we normally capitalize SQL keywords. `s/create table/CREATE TABLE/g`

interleavebench/worker.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (w *worker) run(wg *sync.WaitGroup) {
+	for {
+		start := time.Now()
+
+		// Critical path.
+
+		if _, err := w.db.Exec(`
+		SELECT * FROM
+		merchant JOIN product
```
petermattis: Nit: in my experience, when breaking a query across multiple lines, breaks are placed before SQL keywords. Something like:

```
SELECT * FROM merchant
JOIN product
ON m_id=p_m_id
```

petermattis (approved):
LGTM

Nit on the name. None of our other load generators (which are benchmarks) have `bench` in the name. I'd prefer `s/interleavebench/interleave/g`.

interleavebench/main.go
richardwu: Changed this to retrieve the current user and also take an optional CL argument for the URL.

interleavebench/ddls.go
richardwu: Good point, done.

petermattis (approved):

PR #111, opened 2018-01-09
tpcc: add thinking/keying time and cap workers

jordanlewis:
This commit corrects the tpcc implementation to be to spec with regards
to concurrency and tpmC.

Specifically, all workers now have a keying/thinking time sleep
before/after the actual operation. These sleeps depend on the operation
being performed and are coded to the tpcc spec.

Also, each warehouse now has a maximum of 10 workers as dictated by the
spec. This means that the tpmC of a tpcc load test should be between 8
and 12 tpmC per warehouse, which requires any interesting test with tpcc
to have a lot more data than we've been testing with.

You can disable this new functionality with the -no-wait flag, which
removes the thinking/keying time and cap on concurrency.

### DIFF
--- a/tpcc/worker.go
+++ b/tpcc/worker.go
@@ -19,6 +19,7 @@ import (
 	"database/sql"
 	"fmt"
 	"log"
+	"math"
 	"math/rand"
 	"strconv"
 	"strings"
@@ -31,8 +32,10 @@ import (
 )
 
 type worker struct {
-	db      *sql.DB
-	latency struct {
+	idx       int
+	db        *sql.DB
+	warehouse int
+	latency   struct {
 		sync.Mutex
 		*hdrhistogram.WindowedHistogram
 		byOp []*hdrhistogram.WindowedHistogram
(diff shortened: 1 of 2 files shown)


tpcc/worker.go
```diff
@@ -166,5 +197,11 @@ func (w *worker) run(errCh chan<- error, wg *sync.WaitGroup) {
 		if *maxOps > 0 && v >= *maxOps {
 			return
 		}
+
+		thinkTime := -math.Log(rand.Float64()) * float64(t.thinkTime)
+		if thinkTime > (t.thinkTime * 10) {
+			thinkTime = t.thinkTime * 10
```
petermattis: Where is the `* 10` coming from?

petermattis (approved):
LGTM

tpcc/worker.go
jordanlewis: It's part of the spec. I will add a comment. The think time distribution "may be truncated at 10 times its mean value".

tpcc/worker.go
```diff
@@ -148,7 +172,14 @@ func (w *worker) run(errCh chan<- error, wg *sync.WaitGroup) {
 				break
 			}
 		}
-		if _, err := t.run(w.db, rand.Intn(*warehouses)); err != nil {
+		if *noWait {
+			warehouseID = rand.Intn(*warehouses)
+		}
+
+		time.Sleep(time.Duration(t.keyingTime) * time.Second)
```
rjnn: shouldn't we skip the `time.Sleep` if we're in the `*noWait` case?

rjnn (approved):
LGTM

tpcc/worker.go
jordanlewis: Yes. Thanks for catching that. Done

PR #116, opened 2018-02-07
tpcc: make pg compatible

jordanlewis:
If tpcc is run against a Postgres server (inferred by whether the
database port is 5432), it uses Postgres-compatible schema and set of
queries.

A few things had to change:

1. Postgres doesn't permit descending columns in primary indexes, so we
   got rid of those. I'm not sure if that's to spec.
2. Dropping the database has to be done by dropping the schema instead
   of the database, since you can't drop databases as easily in
   Postgres.
3. Postgres' default transaction isolation is read committed, so a new
   option -serializable was introduced to force serializable mode if
   desired.

### DIFF
--- a/tpcc/ddls.go
+++ b/tpcc/ddls.go
@@ -6,9 +6,10 @@ import (
 )
 
 var ddls = [...]struct {
-	ddl        string
-	interleave string
-	index      bool
+	ddl             string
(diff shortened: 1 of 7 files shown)


nvb (approved):
I tested this out and it's working great!

PR #121, opened 2018-02-15
tpcc: don't wrap errors

jordanlewis:
It breaks the crdb automatic retry library.


nvb:
I think we just need to update our cockroach-go dependency: https://github.com/cockroachdb/cockroach-go/blob/master/crdb/error.go#L5

PR #122, opened 2018-02-15
DNM use RETURNING NOTHING to parallelize statements

nvb:


jordanlewis (approved):
LGTM mod fixing the use of oCID as discussed offline

PR #126, opened 2018-02-16
tpcc: add ramp up time

jordanlewis:


tpcc/main.go
```diff
@@ -43,6 +43,7 @@ var loadIndexes = flag.Bool("load-indexes", false, "Load indexes. Implied by loa
 var maxOps = flag.Uint64("max-ops", 0, "Maximum number of operations to run")
 var noWait = flag.Bool("no-wait", false, "Run in no wait mode (no think/keying time)")
 var opsStats = flag.Bool("ops-stats", false, "Print stats for all operations, not just newOrders")
+var rampMs = flag.Uint64("rampMs", 10000, "Number of milliseconds to ramp up workers over.")
```
petermattis: Does this really need to be milliseconds? Also, you can make this a `flag.Duration`. Lastly, `rampMs` is inconsistent capitalization with the other flags.

tpcc/main.go
```diff
@@ -180,14 +181,17 @@ func main() {
...
 	} else {
 		// 10 workers per warehouse.
-		workers = make([]*worker, *warehouses*10)
-		for i := 0; i < *warehouses; i++ {
-			for j := 0; j < 10; j++ {
-				idx := i*10 + j
-				workers[idx] = newWorker(idx, db, &wg)
-				go workers[idx].run(errCh, i)
+		go func() {
+			workers = make([]*worker, 0, *warehouses*10)
```
petermattis: Since this is happening in a goroutine, any access outside of the goroutine will be racey.

tpcc/main.go
```diff
@@ -180,14 +181,17 @@ func main() {
...
+			for i := 0; i < 10; i++ {
+				for j := 0; j < *warehouses; j++ {
+					idx := i**warehouses + j
+					workers = append(workers, newWorker(idx, db, &wg))
+					time.Sleep(time.Duration(*rampMs/uint64(cap(workers))) * time.Millisecond)
```
petermattis: Is this so all of the workers don't start at the same time? Perhaps it would be better for the workers to start with a randomized think time.

jordanlewis:
Okay, reworked this to just randomize the think time on the first run of all workers.

tpcc/worker.go
```diff
@@ -161,6 +161,7 @@ func newWorker(i int, db *sql.DB, wg *sync.WaitGroup) *worker {
 }
 
 func (w *worker) run(errCh chan<- error, warehouseID int) {
+	firstRun := true
```
petermattis: Nit: you could define this as a loop variable:

```
for firstRun = true; ; {
  ...
}
```

petermattis (approved):
LGTM

jordanlewis:
TFTR

tpcc/worker.go
```diff
@@ -175,7 +175,14 @@ func (w *worker) run(errCh chan<- error, warehouseID int) {
 		if *noWait {
 			warehouseID = rand.Intn(*warehouses)
 		} else {
-			time.Sleep(time.Duration(t.keyingTime) * time.Second)
+			sleepTime := float64(t.keyingTime * 1000)
+			if firstRun {
+				// Sleep for a random duration up to the keying time to smooth out any
+				// potential thundering herd effects when the load generator starts.
+				sleepTime = sleepTime * rand.Float64()
```
rjnn, line 182: Thought: You could use the worker index to smoothly deterministically spread over the ramp time. What you have here (randomness) is smoother than what we have right now, but it's not deterministically smooth, and will cause you frustration when you look at experiment after experiment, and each one takes a varying amount of time to settle down.

rjnn (approved):
LGTM. I think there's A Better Way™, but I'll be happy if you merge it as-is too.

tpcc/worker.go
jordanlewis: That's what I had at first. Peter suggested replacing that with this simpler method, which prevents the proximal issue of x-treme contention right at the start of the test due to the fact that the workers were all starting in sequence.

rjnn: Gah. Reviews in github over rebases are hard. I should have checked. Apologies.

I disagree with Peter, but by waiting over the long weekend I've given up my right to complain. :)

:shipit:!

PR #128, opened 2018-02-27
tpcc: use lookup join in stock level

jordanlewis:


rjnn (approved):
LGTM

tpcc/stock_level.go
```diff
@@ -57,6 +57,12 @@ func (s stockLevel) run(db *sql.DB, wID int) (interface{}, error) {
 		dID:       rand.Intn(9) + 1,
 	}
 
+	// This is the only join in the application, so we don't need to worry about
+	// this setting persisting incorrectly across queries.
+	if _, err := db.Exec(`set experimental_force_lookup_join=true`); err != nil {
```
petermattis, line 62: This is not doing what you think it is doing. This is setting a session-level option on a single connection and we need to set this option on every connection. I think you need to move this inside the `crdb.ExecuteTx` and use `tx.Exec` instead.

rjnn: But this is being called inside `stockLevel.run`. It's not being done transactionally, but it is being done every time before the join is run. Am I missing something to do with the multiplexing of connections internally in `db`?

petermattis: There is no guarantee that the connection used for the `db.Exec` is the same one used for the transaction below.

PR #131, opened 2018-02-28
tpcc: batch all writes in newOrder txn

nvb:
This change removes the iteration over each item an order in the `newOrder` transaction. It replaces
this iteration with batched operations that each work over all items in the order. This drastically reduces the number of statements that are run in the transaction and avoids O(item_count) round trips.

### DIFF
--- a/tpcc/new_order.go
+++ b/tpcc/new_order.go
@@ -19,6 +19,7 @@ import (
 	"database/sql"
 	"fmt"
 	"math/rand"
+	"sort"
 	"strings"
 	"time"
 
@@ -42,6 +43,7 @@ import (
 type orderItem struct {
 	olSupplyWID  int    // supplying warehouse id
 	olIID        int    // item id
+	olNumber     int    // item number in order
 	iName        string // item name
 	olQuantity   int    // order quantity
 	sQuantity    int    // stock quantity
@@ -86,6 +88,11 @@ func (n newOrder) run(db *sql.DB, wID int) (interface{}, error) {
 	}
 	d.items = make([]orderItem, d.oOlCnt)
 
+	// itemIDs tracks the item ids in the order so that we can prevent adding
+	// multiple items with the same ID. This would not make sense because each
+	// orderItem already tracks a quantity that can be larger than 1.
+	itemIDs := make(map[int]struct{})
+
 	// 2.4.1.4: A fixed 1% of the New-Order transactions are chosen at random to
 	// simulate user data entry errors and exercise the performance of rolling
 	// back update transactions.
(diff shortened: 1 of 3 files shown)


tpcc/new_order.go
```diff
@@ -155,88 +166,165 @@ func (n newOrder) run(db *sql.DB, wID int) (interface{}, error) {
...
+			for i, item := range d.items {
+				itemIDs[i] = item.olIID
+				itemIDPls[i] = fmt.Sprintf("$%d", i+1)
 			}
-
-			selectItem, err := tx.Prepare(`SELECT i_price, i_name, i_data FROM item WHERE i_id=$1`)
+			rows, err := tx.Query(fmt.Sprintf(`
+				SELECT i_price, i_name, i_data 
+				FROM item 
+				WHERE i_id IN (%s)`, strings.Join(itemIDPls, ", ")),
```
jordanlewis: I feel like you might as well just interpolate the items right into the query, like Peter was saying. But we can make that a separate patch.

tpcc/new_order.go
```diff
@@ -155,88 +166,165 @@ func (n newOrder) run(db *sql.DB, wID int) (interface{}, error) {
...
+			)
 			if err != nil {
 				return err
 			}
+			distInfos := make([]string, d.oOlCnt)
+			sQuantityUpdateCases := make([]string, d.oOlCnt)
+			sYtdUpdateCases := make([]string, d.oOlCnt)
+			sOrderCntUpdateCases := make([]string, d.oOlCnt)
+			sRemoteCntUpdateCases := make([]string, d.oOlCnt)
+			for i := 0; rows.Next(); i++ {
```
jordanlewis: I'm nervous that this is broken because the select doesn't have an order by... don't we need the results to be in the same order that we asked for from d.items?

nvb: Yeah, you're completely right. I meant to fix this when I made the change to ensure that all items are unique.

nvb: Done.

jordanlewis (approved):
LGTM but do note that adding an ORDER BY is not something that the spec asks you to do. The spec does seem flexible with regards to what actually gets executed, but I do just want to note that I'm nervous about diverging so much from what the other implementations do.

petermattis:
We don't actually need the `ORDER BY`. We could parse each result row and store the data for the correct item. The `ORDER BY` just makes that easier.

nvb:
Yeah the `ORDER BY` isn't strictly needed here, it just makes things a lot easier and cleaner client-side.

PR #132, opened 2018-02-28
Always bind workers to warehouses

petermattis:
Change the behavior when `-no-wait` is specified to assign each worker
to a warehouse. Previously each worker was selecting a random warehouse
to operate on. This allows experimentation with contention. For example,
`-warehouses=1 -concurrency=1 -no-wait` shows the throughput of a single
worker without contention. Specify higher values of concurrency to see
more contention. And we can also specify `-warehouses=10
-concurrency=10` to see scalability in the absence of contention.

Change the ramp-up behavior to factor in the think time as well. This is
an attempt to reduce the thundering herd effect we're seeing when the
load generator starts.

### DIFF
--- a/tpcc/worker.go
+++ b/tpcc/worker.go
@@ -76,6 +76,18 @@ type tx struct {
 	thinkTime  float64 // minimum mean of think time distribution, 5.2.5.7
 }
 
+func (t tx) randThinkTime() time.Duration {
+	// 5.2.5.4: Think time is taken independently from a negative exponential
+	// distribution. Think time = -log(r) * u, where r is a uniform random number
+	// between 0 and 1 and u is the mean think time per operation.
+	// Each distribution is truncated at 10 times its mean value.
+	thinkTime := -math.Log(rand.Float64()) * float64(t.thinkTime)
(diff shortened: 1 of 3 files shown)


petermattis:
This is ready for a look.

nvb (approved):

tpcc/worker.go
```diff
@@ -172,17 +184,15 @@ func (w *worker) run(errCh chan<- error, warehouseID int) {
...
-			warehouseID = rand.Intn(*warehouses)
-		} else {
-			sleepTime := float64(t.keyingTime * 1000)
+		if !*noWait {
+			sleepTime := time.Duration(t.keyingTime) * time.Second
 			if firstRun {
 				// Sleep for a random duration up to the keying time to smooth out any
 				// potential thundering herd effects when the load generator starts.
-				sleepTime = sleepTime * rand.Float64()
+				sleepTime = time.Duration(float64(sleepTime)*rand.Float64()) + t.randThinkTime()/2
```
jordanlewis, line 192: this is weird but harmless. why sleep a randThinkTime during the keying time in nowait?

jordanlewis (commented):
LGTM

PR #133, opened 2018-02-28
tpcc: shard warehouses by machine

jordanlewis:
The load generator now accepts a list of machine urls, and evenly assigns each warehouse to a machine.

Also, pin the number of client connections to the number of warehouses for simplicity's sake.


petermattis (approved):
LGTM

I wanted to build upon this, so I'm going to merge it for you.

PR #135, opened 2018-03-01
tpcc: break up delivery into n txns

jordanlewis:
The spec allows us to do the delivery business txn in up to 10 SQL txns. We were previously doing it in one giant SQL txn - now we do 10, which should have less contention.

Also, use string interpolation instead of prepared statements and coalesce two queries by using a subquery.

### DIFF
--- a/tpcc/delivery.go
+++ b/tpcc/delivery.go
@@ -92,19 +92,15 @@ func (del delivery) run(db *sql.DB, wID int) (interface{}, error) {
 					olDeliveryD.Format("2006-01-02 15:04:05"), wID, dID, oID)); err != nil {
 					return err
 				}
-				var olTotal float64
-				if err := tx.QueryRow(fmt.Sprintf(`
-						SELECT SUM(ol_amount) FROM order_line
-						WHERE ol_w_id = %d AND ol_d_id = %d AND ol_o_id = %d`,
-					wID, dID, oID)).Scan(&olTotal); err != nil {
-					return err
-				}
 				if _, err := tx.Exec(fmt.Sprintf(`
 						UPDATE customer
 						SET (c_balance, c_delivery_cnt) =
-							(c_balance + %f, c_delivery_cnt + 1)
+							(c_balance + (
+								SELECT SUM(ol_amount) FROM order_line
+								WHERE ol_w_id = %d AND ol_d_id = %d AND ol_o_id = %d
+							), c_delivery_cnt + 1)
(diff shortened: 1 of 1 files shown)


tpcc/delivery.go
```diff
@@ -48,87 +49,64 @@ func (del delivery) run(db *sql.DB, wID int) (interface{}, error) {
...
 					return err
 				}
-				if _, err := updateCustomer.Exec(olTotal, wID, dID, oCID); err != nil {
+				if _, err := tx.Exec(fmt.Sprintf(`
+						UPDATE customer
+						SET (c_balance, c_delivery_cnt) =
+							(c_balance + (
+								SELECT SUM(ol_amount) FROM order_line
+								WHERE ol_w_id = %d AND ol_d_id = %d AND ol_o_id = %d
+							), c_delivery_cnt + 1)
```
petermattis, line 101: I think we can avoid the subquery here by having the `UPDATE order_line` return `ol_amount` and doing the sum on the client side. Or perhaps something like `SELECT SUM(ol_amount) FROM [UPDATE order_line ... RETURNING ol_amount]`. Not sure if that works, though.

tpcc/delivery.go
```diff
@@ -48,87 +49,64 @@ func (del delivery) run(db *sql.DB, wID int) (interface{}, error) {
...
 						return err
 					}
-					continue
+					return nil
 				}
-				if _, err := delNewOrder.Exec(wID, dID, oID); err != nil {
+				if _, err := tx.Exec(fmt.Sprintf(`
+						DELETE FROM new_order
+						WHERE no_w_id = %d AND no_d_id = %d AND no_o_id = %d`,
+					wID, dID, oID)); err != nil {
```
petermattis, line 76: I think we can combine the `DELETE FROM new_order` with the previous `SELECT` using `DELETE FROM new_order WHERE no_w_id = %d AND no_d_id = %d ORDER BY no_o_id ASC LIMIT 1 RETURNING  no_o_id`. Would definitely need to test this.

nvb:
This is a clear win compared to the previous approach of a single transaction with an internal loop, but I think a batching approach would still be preferable for two reasons. The first is that we're seeing CPU saturation at the boundaries of our performance. Batching would help in this regard because it would reduce the number of statements processed and the number of network round trips required by this transaction by roughly a factor of 10. Batching also allows us to gain parallelism between each district, which will reduce the aggregate latency of this operation significantly.

PR #138, opened 2018-03-01
tpcc: fix payment

### LINKED ISSUE #134: tpcc: payment is missing 2 steps
I just noticed that the implementation of the payment transaction is incomplete. It's supposed to update the `c_data` field if the customer had bad credit. Instead, it doesn't. I'll fix this.

jordanlewis:
It now updates c_data when the customer has bad credit. Before, it
didn't.

Closes #134.

### DIFF
--- a/tpcc/payment.go
+++ b/tpcc/payment.go
@@ -175,29 +175,28 @@ func (p payment) run(db *sql.DB, wID int) (interface{}, error) {
 			}
 
 			// Update customer with payment.
+			// If the customer has bad credit, update the customer's C_DATA and return
+			// the first 200 characters of it, which is supposed to get displayed by
+			// the terminal. See 2.5.3.3 and 2.5.2.2.
 			if err := tx.QueryRow(`
 				UPDATE customer
-				SET (c_balance, c_ytd_payment, c_payment_cnt) =
-					(c_balance - $1, c_ytd_payment + $1, c_payment_cnt + 1)
+				SET (c_balance, c_ytd_payment, c_payment_cnt, c_data) =
+					(c_balance - $1, c_ytd_payment + $1, c_payment_cnt + 1,
+					 case c_credit when 'BC' then
+					 left(c_id::text || c_d_id::text || c_w_id::text || $5::text || $6::text || $1::text || c_data, 500)
+					 else c_data end)
 				WHERE c_w_id = $2 AND c_d_id = $3 AND c_id = $4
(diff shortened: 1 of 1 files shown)


tpcc/payment.go
```diff
@@ -175,29 +175,28 @@ func (p payment) run(db *sql.DB, wID int) (interface{}, error) {
 			}
 
 			// Update customer with payment.
+			// If the customer has bad credit, update the customer's C_DATA and return
+			// the first 200 characters of it, which is supposed to get displayed by
+			// the terminal. See 2.5.3.3.
```
nvb: > See 2.5.3.3

2.5.2.2?

tpcc/payment.go
```diff
@@ -175,29 +175,28 @@ func (p payment) run(db *sql.DB, wID int) (interface{}, error) {
...
 			if err := tx.QueryRow(`
 				UPDATE customer
-				SET (c_balance, c_ytd_payment, c_payment_cnt) =
-					(c_balance - $1, c_ytd_payment + $1, c_payment_cnt + 1)
+				SET (c_balance, c_ytd_payment, c_payment_cnt, c_data) =
+					(c_balance - $1, c_ytd_payment + $1, c_payment_cnt + 1,
+					 case c_credit when 'BC' then
+					 left(c_id::text || c_d_id::text || c_w_id::text || $5::text || $6::text || $1::text || c_data, 500)
```
nvb, line 186: Is there supposed to be a pipe between `hAmount` and `cData` like we had before?

nvb (approved):

tpcc/payment.go
jordanlewis: 2.5.3.3 has the part that says "just return 200 characters", which is what i was getting at.

tpcc/payment.go
jordanlewis: No, I don't know where I got that from. I must have copied another benchmark. The spec says:

> If the value of C_CREDIT is equal to "BC", then C_DATA is also retrieved from the selected customer and the following history information: C_ID, C_D_ID, C_W_ID, D_ID, W_ID, and H_AMOUNT, are inserted at the left of the C_DATA field by shifting the existing content of C_DATA to the right by an equal number of bytes and by discarding the bytes that are shifted out of the right side of the C_DATA field. The content of the C_DATA field never exceeds 500 characters.

PR #139, opened 2018-03-01
tpcc: add -split and -scatter flags

petermattis:


tpcc/main.go
```diff
@@ -193,24 +203,22 @@ func main() {
...
 	errCh := make(chan error)
 	var wg sync.WaitGroup
-	if *noWait {
+	for i := range workers {
+		w := i % *warehouses
+		workers[i] = newWorker(i, w, dbs[w%len(dbs)], &wg)
+	}
+
+	go func() {
+		// Starter the workers evenly spaced over 30s to avoid any thundering
```
jordanlewis, line 213: Hah! I told you this was a good idea.

jordanlewis: We should make this configurable with a flag.

tpcc/worker.go
```diff
@@ -189,14 +189,14 @@ func (w *worker) run(errCh chan<- error, warehouseID int) {
 			if firstRun {
 				// Sleep for a random duration up to the keying time to smooth out any
 				// potential thundering herd effects when the load generator starts.
-				sleepTime = time.Duration(float64(sleepTime)*rand.Float64()) + t.randThinkTime()/2
+				sleepTime = time.Duration(float64(sleepTime) * rand.Float64())
```
jordanlewis, line 194: Can we ditch this `firstRun` stuff now that there's a ramp up time?

jordanlewis (approved):

tpcc/main.go
petermattis: Added a `-ramp` flag.

tpcc/worker.go
petermattis: I've added a TODO to revisit this. I don't want to rip it out without testing and I won't be able to test until later tonight or tomorrow.

jordanlewis (approved):
LGTM

tpcc/ddls.go
```diff
@@ -264,6 +264,30 @@ func loadSchema(db *sql.DB, interleave bool, index bool, usePostgres bool) {
 	fmt.Printf("\n")
 }
 
+func splitTables(db *sql.DB, warehouses int) {
```
nvb: Consider adding a note here that because we always split at exactly the same point, running with `-split` is idempotent so it's safe to always use (unlike `-load`).

nvb: Just missed your merge! Feel free to ignore.

petermattis: I'll send a follow-on PR.

PR #141, opened 2018-03-01
tpcc: fix "sum(O_OL_CNT) = # of rows in the ORDER-LINE table" check

nvb:
The check was trying to scan into a value instead of a pointer, which
was throwing the error
```
check consistency failed: sql: Scan error on column index 0: destination not a pointer
```

It was also comparing sum(o_ol_cnt) with the number of rows in the new_order table for
a district instead of the number of rows in the order_line table for that district.

### DIFF
--- a/tpcc/check.go
+++ b/tpcc/check.go
@@ -20,7 +20,6 @@ func checkConsistency(db *sql.DB) error {
 	// Reused checks
 
 	maxNewOIDs := make([]int, *warehouses*10)
-	countNewOrders := make([]int, *warehouses*10)
 
 	sliceIdx := func(wID, dID int) int {
 		return wID*10 + dID - 1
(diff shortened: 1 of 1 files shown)


jordanlewis (approved):
Whoops. It's possible that I just committed this without actually finishing it, so you might want to double check that that particular check even works/makes sense.

nvb:
It's failing at the moment, so I hope it's just incomplete and not revealing a larger issue.

nvb:
Oh, I think that's because it's comparing `sum(o_ol_cnt)` with the number of rows in the **new_order** table for a district instead of the number of rows in the **order_line** table for a district! I'll fix that in a second commit.

nvb:
PTAL. With this change, all checks pass!

tpcc/check.go
```diff
@@ -93,16 +91,21 @@ func checkConsistency(db *sql.DB) error {
 
 	// 3.3.2.4
 	// sum(O_OL_CNT) = # of rows in the ORDER-LINE table for each warehouse/district
-	var sumOOLCnt int
+	var sumOOLCnt, countNewOrderLint int
```
jordanlewis: why is this called Lint? just curious

jordanlewis (approved):
LGTM thanks!

tpcc/check.go
nvb: Because I'm having trouble spelling "line" today 😄

jordanlewis: Oh, I get it, haha

PR #142, opened 2018-03-02
tpcc: use deck method of regulating txn mix

jordanlewis:
The naive randomness used originally (picking randomly from the
transaction mix distribution without hysteris) leads to a potentially
invalid transaction mix.

The spec suggests using a deck of transactions to regulate the mix,
where the deck is iterated through in random order and shuffled when
exhausted. Now, we use that method.

Note that the default transaction mix changed slightly. The spec isn't
precise on what the transaction mix is, but it clearly says that the
deck method can be implemented with 10 cards for each of payment and new
order, and 1 each of the rest, so that's what the mix is now.

### DIFF
--- a/teamcity-push.sh
+++ b/teamcity-push.sh
@@ -48,7 +48,7 @@ for proj in kv ycsb tpch tpcc ; do
         --workdir=/go/src/github.com/cockroachdb/loadgen \
         --volume="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)":/go/src/github.com/cockroachdb/loadgen \
         --rm \
-        cockroachdb/builder:20170422-212842 make ${proj} STATIC=1
+        cockroachdb/builder:20180220-200046 make ${proj} STATIC=1
     strip -S ${proj}/${proj}
     push_one_binary ${proj}/${proj}
 done
--- a/teamcity-test.sh
+++ b/teamcity-test.sh
@@ -8,4 +8,4 @@ docker run \
     --volume="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)":/go/src/github.com/cockroachdb/loadgen \
     --volume="/tmp/loadgenbin":/go/src/github.com/cockroachdb/loadgen/bin \
     --rm \
-    cockroachdb/builder:20170422-212842 make all | go-test-teamcity
+    cockroachdb/builder:20180220-200046 make all | go-test-teamcity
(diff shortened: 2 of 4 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[tpcc/worker.go, line 195 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/142#-L6b7CuRHFChfEVGtsG2:-L6b7CuSnJX0Rdz6cDDA:bor3wpg) ([raw file](https://github.com/cockroachdb/loadgen/blob/051db52ca849ca1e9702b8e72ce90130b4929cc7/tpcc/worker.go#L195)):*
> ```Go
> 		// uniformly random order.
> 		if permIdx == len(deck) {
> 			deckPerm = rand.Perm(len(deck))
> ```

I was imagining you would use `rand.Shuffle` here. In order to do so, you'll need to make a copy of the deck for each worker.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/142#-:-L6b7QEPZKIY4C9kBhMR:bnfp4nl)*

rjnn:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/142#-:-L6b8nzzpcLLkl2uCVVj:bnfp4nl)*

jordanlewis:
Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[tpcc/worker.go, line 195 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/142#-L6b7CuRHFChfEVGtsG2:-L6bLHDcd8Lg2sbzpXDa:b-eghnsz) ([raw file](https://github.com/cockroachdb/loadgen/blob/051db52ca849ca1e9702b8e72ce90130b4929cc7/tpcc/worker.go#L195)):*

I was imagining you would use `rand.Shuffle` here. In order to do so, you'll need to make a copy of the deck for each worker.
</blockquote></details>

Done. I used `rand.Perm` because I didn't realize they added `rand.Shuffle` (in 1.10).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/142)*

petermattis:
:shipit: 

---

Review status: 1 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/142#-:-L6bM0aeexucZ26eRBbR:bizlbj6)*

PR #144, opened 2018-03-02
tpcc: display p90 latencies

petermattis:
Display p90 latencies and display latencies for all op types when the
test ends.

### DIFF
--- a/tpcc/main.go
+++ b/tpcc/main.go
@@ -292,17 +292,18 @@ func main() {
 			elapsed := now.Sub(lastNow)
 			ops := numOps
 			if i%20 == 0 {
-				fmt.Println("_time______opName__ops/s(inst)__ops/s(cum)__p50(ms)__p95(ms)__p99(ms)_pMax(ms)")
+				fmt.Println("_time______opName__ops/s(inst)__ops/s(cum)__p50(ms)__p90(ms)__p95(ms)__p99(ms)_pMax(ms)")
 			}
 			i++
 			totalTime := time.Duration(time.Since(start).Seconds()+0.5) * time.Second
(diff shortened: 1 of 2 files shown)


tpcc/main.go
```diff
@@ -337,17 +339,22 @@ func main() {
...
-				time.Duration(cumLatencyByOp[newOrderType].Mean()).Seconds()*1000,
-				time.Duration(cumLatencyByOp[newOrderType].ValueAtQuantile(50)).Seconds()*1000,
-				time.Duration(cumLatencyByOp[newOrderType].ValueAtQuantile(95)).Seconds()*1000,
-				time.Duration(cumLatencyByOp[newOrderType].ValueAtQuantile(99)).Seconds()*1000,
-				time.Duration(cumLatencyByOp[newOrderType].ValueAtQuantile(100)).Seconds()*1000)
+			fmt.Println("\n_elapsed______opName__ops(total)__ops/m(cum)__avg(ms)__p50(ms)__p90(ms)__p95(ms)__p99(ms)_pMax(ms)")
+			totalTime := time.Since(start).Seconds()
+			for i, h := range cumLatencyByOp {
+				ops := atomic.LoadUint64(&txs[i].numOps)
+				elapsed := time.Since(start).Seconds()
```
jordanlewis: NYC but just noticed that this should be pulled outside of the loop or else each op will get a slightly different elapsed time.

jordanlewis (approved):

tpcc/main.go
petermattis: Heh, this is the same value as `totalTime`. Fixed.

PR #145, opened 2018-03-02
tpcc: remove the firstRun sleep randomization

petermattis:
The `-ramp` mechanism is sufficient to smooth out the initial load.


jordanlewis:
LGTM.

PR #137, opened 2018-03-01
tpcc: batch all writes in delivery txn

nvb:
Alternative to #135.

This change replaces the for-each-district loop in the delivery
transaction with a series of batched statements that perform all
writes together. This reduces the number of round-trips required
by the transaction by an order of magnitude.

PR #146, opened 2018-03-02
tpcc: deferred mode for delivery txn

jordanlewis:


petermattis:
Per offline discussion, the default of 1 deferred worker seems to be unable to keep up with the `delivery` ops. Tuning the number of deferred workers could be tricky. Or we could have a fancier mechanism. In the short term #137 seems like the best approach to reduce the impact of `delivery` ops.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/146#-:-L6bxYTErMtVxLXdQh2T:b-fgb96m)*

PR #149, opened 2018-03-03
tpcc: fix random customer selection method in payment

petermattis:
The previous code was selecting customers by last name with 66%
probability, while the spec requires 60% probability.


jordanlewis (approved):
Oops, nice catch. I wonder where I got Intn(9) from, that doesn't make much sense.

PR #152, opened 2018-03-03
tpcc: permit loading a warehouse subset

jordanlewis:
Introduce new flags and functionality for improved parallel loading.

  -load-offset starts loading warehouses at a specified offset
  -load-indexes controls whether indexes will be built
  -load-schema controls whether schema will be built
  -conns-per-url controls the number of connections per url. if
    unspecified, the previous default of nWarehouses is kept.

Also improves the appearance of the loading process to make it more
clear what the current phase is.

### DIFF
--- a/tpcc/ddls.go
+++ b/tpcc/ddls.go
@@ -234,8 +234,8 @@ create table order_line (
 	},
 }
 
-// loadSchema loads the entire TPCC schema into the database.
-func loadSchema(db *sql.DB, interleave bool, index bool, usePostgres bool) {
+// doLoadSchema loads the entire TPCC schema into the database.
+func doLoadSchema(db *sql.DB, interleave bool, index bool, usePostgres bool) {
 	schemaType := "tables"
 	if index {
 		schemaType = "indexes"
--- a/tpcc/generate.go
+++ b/tpcc/generate.go
@@ -57,7 +57,7 @@ func prepare(db *sql.DB, query string) *sql.Stmt {
 	return stmt
 }
 
-func parallelLoad(n int, batchSize int, entityName string, loader func(int, int)) {
+func parallelLoad(n int, batchSize int, entityName string, depth int, loader func(int, int)) {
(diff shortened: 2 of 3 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[tpcc/generate.go, line 121 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/152#-L6hACYNuDHn0PVX9uda:-L6hACYNuDHn0PVX9udb:b-zaf7yi) ([raw file](https://github.com/cockroachdb/loadgen/blob/4c15831e476064ef6f1ea527db3940430e85ab10/tpcc/generate.go#L121)):*
> ```Go
> 	}
> 	if nItems == 0 {
> 		// Only load items if we haven't loaded them already.
> ```

This seems racey if you have multiple `tpcc` processes trying to load items in parallel. Perhaps there should be an explicit flag to go along with the check.

---

*[tpcc/main.go, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/152#-L6hgZRtI3p6I0WkoDqt:-L6hgZRtI3p6I0WkoDqu:b-psnvgm) ([raw file](https://github.com/cockroachdb/loadgen/blob/4c15831e476064ef6f1ea527db3940430e85ab10/tpcc/main.go#L38)):*
> ```Go
> var check = flag.Bool("check", false, "Run consistency checks.")
> var concurrency = flag.Int("concurrency", 2*runtime.NumCPU(), "Number of terminals (ignored unless -no-wait is specified)")
> var connsPerURL = flag.Int("conns-per-url", 0, "Number of connections to open per url")
> ```

Add a parenthetical: `(0 defaults to -warehouses)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/152#-:-L6hgtdC-zj4-RD2q3RQ:bnfp4nl)*

PR #156, opened 2018-03-05
tpcc: add -partition flag

petermattis:
Partition all of the tables and indexes. Most of the tables are
partitioned by warehouse ID. The item table is partitioned by item
ID. The primary index of the history table is partitioned by
rowid (random 63-bit positive integer) while the other history table
indexes are partitioned by warehouse ID.

The motivation here is to be able to spread the partitions evenly in a
cluster in order to achieve better range distribution. This will require
setting zone configs for all of these partitions and starting up the
cluster with attributes or localities on each node.

### DIFF
--- a/tpcc/main.go
+++ b/tpcc/main.go
@@ -45,6 +45,7 @@ var noWait = flag.Bool("no-wait", false, "Run in no wait mode (no think/keying t
 var opsStats = flag.Bool("ops-stats", false, "Print stats for all operations, not just newOrders")
 var scatter = flag.Bool("scatter", false, "Scatter ranges")
 var split = flag.Bool("split", false, "Split tables")
+var partitions = flag.Int("partitions", 0, "Partition tables")
 var run = flag.Bool("run", true, "Run benchmark")
 var ramp = flag.Duration("ramp", 30*time.Second, "The duration over which to ramp up workers")
 var tolerateErrors = flag.Bool("tolerate-errors", false, "Keep running on error")
@@ -181,6 +182,10 @@ func main() {
 
 	if *split {
 		splitTables(db, *warehouses)
+	}
+
+	if *partitions > 0 {
+		partitionTables(db, *warehouses, *partitions)
 	}
 
 	if *scatter {
--- a/tpcc/partition.go
+++ b/tpcc/partition.go
@@ -7,12 +7,21 @@ import (
 	"math"
 )
 
-func partitionWarehouse(db *sql.DB, wIDs []int) {
+func configureZone(db *sql.DB, table, partition string, constraint int) {
+	sql := fmt.Sprintf(
+		`ALTER PARTITION %s OF TABLE %s EXPERIMENTAL CONFIGURE ZONE 'constraints: [+rack=%d]'`,
+		partition, table, constraint)
+	if _, err := db.Exec(sql); err != nil {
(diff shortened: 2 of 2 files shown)


danhhz:
:lgtm: cool! this is definitely the most interesting partitioning we've ever tested

there's a lot of repetition here and i didn't look closely for copy/paste errors but seems like it's doing the right thing

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/156#-:-L6r2FC046iDQdFWLGy1:bgg7ies)*

jordanlewis:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[tpcc/partition.go, line 198 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/156#-L6r4RzSuenayKkSMo4z:-L6r4RzSuenayKkSMo5-:b-7f4akn) ([raw file](https://github.com/cockroachdb/loadgen/blob/f9a2ce94296b81811f058c2962688923de07a511/tpcc/partition.go#L198)):*
> ```Go
> }
> 
> func partitionItem(db *sql.DB, partitions int) {
> ```

The item table is read only. Maybe what we should do is make one copy of the item table per node, and change its zone config to have rf=1. Is that too hacky?

Haha just searched the spec for "replication" and found this gem:

```
Comment 2: If the implementation uses replication on the ITEM table and all transactions in Isolation Test 7 use the same copy of the ITEM table, updates to the ITEM table are not required to be propagated to other copies of the ITEM table. This relaxation of ACID properties on a replicated table is only valid under the above conditions and in the context of Isolation Test 7.
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/156#-:-L6r56kOhTcv_zxFZKVa:bnfp4nl)*

petermattis:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[tpcc/partition.go, line 198 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/156#-L6r4RzSuenayKkSMo4z:-L6rJUunDaW7i8N04gCo:b-s8clme) ([raw file](https://github.com/cockroachdb/loadgen/blob/f9a2ce94296b81811f058c2962688923de07a511/tpcc/partition.go#L198)):*

The item table is read only. Maybe what we should do is make one copy of the item table per node, and change its zone config to have rf=1. Is that too hacky?

Haha just searched the spec for "replication" and found this gem:

```
Comment 2: If the implementation uses replication on the ITEM table and all transactions in Isolation Test 7 use the same copy of the ITEM table, updates to the ITEM table are not required to be propagated to other copies of the ITEM table. This relaxation of ACID properties on a replicated table is only valid under the above conditions and in the context of Isolation Test 7.
```
</blockquote></details>

Yeah, I noticed that the `item` table is never modified. I don't think there is anything easy for us to do in the short term. When the max-safe-timestamp arrives, we can query at a historical timestamp which will allow reading from followers. And then we can increase the replication of the item table to provide more readable copies. Hopefully the work in 2.1 will allows this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/156)*

PR #158, opened 2018-03-06
tpcc: partition the stock_s_i_id_idx index and split the history table

petermattis:


rjnn:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/loadgen/158)*

petermattis:
@arjunravinarayan Was that meant to be an LGTM?

rjnn:
No I intentionally did not LGTM this as I did not understand the purpose of this splitting, or the underlying data distribution (and thus whether this partitioning would achieve the required range spreading). However, after our offline discussion, I now understand what's going on.

LGTM.

rjnn (approved):

tbg:
> I now understand what's going on.

Great, can you share? 🍿

petermattis:
> Great, can you share? 🍿

Failing to partition the stock index was just an oversight of the initial partitioning work.

Splitting of the history table is necessary to avoid hotspots in writing history entries. Note that we manually set the `history.rowid` column to a random 63-bit integer, but we also need to have a sufficient number of ranges in the history table and have those ranges at the correct splits. Previously, the history table had a large number of ranges, but they were all clustered in a single partition because the TPCC dataset generated by `workload` have small contiguous `rowid` values. I need to file an issue about that.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/158#-:-L70UgsN-5V-CKxEAoIE:buctnwj)*

rjnn:
Two things being split here: the history table and the stocks@s_i_id_idx index.

The history table has a uint64 primary key that is randomly generated uniformly at random across the keyspace. When a txn is processed, a new row is added to the history table. This splitting ensures that the write pressure on the history table is spread evenly across the cluster.

The index splitting is just the the index was overlooked when  splitting the stocks table.

PR #161, opened 2018-03-08
tpcc: refactor checks 1 through 5

rjnn:
On 100 warehouses, the checks now take:

3.3.2.1: 100 rows, passed in 0.061173 seconds.
3.3.2.2: 1000 rows, passed in 4.662491 seconds.
3.3.2.3: 1000 rows, passed in 1.034691 seconds.
3.3.2.4: 1000 rows, passed in 30.794494 seconds.
3.3.2.5: passed in 21.333984 seconds.

### DIFF
--- a/tpcc/check.go
+++ b/tpcc/check.go
@@ -3,129 +3,274 @@ package main
 import (
 	"database/sql"
 	"fmt"
+
+	"time"
+
+	"github.com/pkg/errors"
 )
 
-func forEachWarehouseAndDistrict(fn func(wID, dID int) error) error {
-	for wID := 0; wID < *warehouses; wID++ {
-		for dID := 1; dID < 10; dID++ {
-			if err := fn(wID, dID); err != nil {
-				return err
-			}
-		}
+func check3321(db *sql.DB) error {
+	// 3.3.2.1 Entries in the WAREHOUSE and DISTRICT tables must satisfy the relationship:
+	// W_YTD = sum (D_YTD)
+
+	row := db.QueryRow(`
+SELECT COUNT(*) 
+FROM warehouse 
+FULL OUTER JOIN 
+(SELECT d_w_id, SUM(d_ytd) as sum_d_ytd FROM district GROUP BY d_w_id) 
+ON (w_id = d_w_id)
+WHERE w_ytd != sum_d_ytd
+`)
+
+	var i int
+	if err := row.Scan(&i); err != nil {
+		return err
 	}
+
+	if i != 0 {
+		return errors.Errorf("%d rows returned, expected zero", i)
+	}
+
 	return nil
 }
 
-func checkConsistency(db *sql.DB) error {
-	// Reused checks
+func check3322(db *sql.DB) error {
+	// Entries in the DISTRICT, ORDER, and NEW-ORDER tables must satisfy the relationship:
+	// D_NEXT_O_ID - 1 = max(O_ID) = max(NO_O_ID)
 
-	maxNewOIDs := make([]int, *warehouses*10)
-
-	sliceIdx := func(wID, dID int) int {
-		return wID*10 + dID - 1
+	districtRows, err := db.Query("SELECT d_next_o_id FROM district ORDER BY d_w_id, d_id")
+	if err != nil {
+		return err
 	}
-
-	// 3.3.2.1
-	// W_YTD = sum(D_YTD) for all warehouses and districts
-	var wYTD, sumDYTD float64
-	for i := 0; i < *warehouses; i++ {
-		if err := db.QueryRow("SELECT w_ytd FROM warehouse WHERE w_id=$1", i).Scan(&wYTD); err != nil {
-			return err
-		}
-		if err := db.QueryRow("SELECT SUM(d_ytd) FROM district WHERE d_w_id=$1", i).Scan(&sumDYTD); err != nil {
-			return err
-		}
-		if wYTD != sumDYTD {
-			fmt.Printf("check failed: w_ytd=%f != sum(d_ytd)=%f for warehouse %d\n", wYTD, sumDYTD, i)
-		}
+	newOrderRows, err := db.Query(`
+SELECT MAX(no_o_id) FROM tpcc.new_order GROUP BY no_d_id, no_w_id ORDER BY no_w_id, no_d_id
+`)
+	if err != nil {
+		return err
 	}
-
-	// 3.3.2.2
-	// D_NEXT_O_ID - 1 = max(O_ID) = max(NO_O_ID) for all districts
-	var dNextOID, maxOID, maxNewOID int
-	if err := forEachWarehouseAndDistrict(func(wID, dID int) error {
-		if err := db.QueryRow(`SELECT d_next_o_id FROM district WHERE d_w_id=$1 AND d_id=$2`,
-			wID, dID,
-		).Scan(&dNextOID); err != nil {
-			return err
-		}
-		if err := db.QueryRow(`SELECT max(o_id) FROM "order" WHERE o_w_id=$1 AND o_d_id=$2`,
-			wID, dID,
-		).Scan(&maxOID); err != nil {
-			return err
-		}
-		if err := db.QueryRow(`SELECT max(no_o_id) FROM new_order WHERE no_w_id=$1 AND no_d_id=$2`,
-			wID, dID,
-		).Scan(&maxNewOID); err != nil {
-			return err
-		}
-		maxNewOIDs[sliceIdx(wID, dID)] = maxNewOID
-
-		if dNextOID != maxOID+1 || dNextOID != maxNewOID+1 {
-			fmt.Printf("check failed: d_next_o_id=%d != max(o_id)+1=%d != max(no_o_id)+1=%d for warehouse %d district %d\n",
-				dNextOID, maxOID, maxNewOID, wID, dID)
-		}
-		return nil
-	}); err != nil {
+	orderRows, err := db.Query(`
+SELECT MAX(o_id) FROM tpcc.order GROUP BY o_d_id, o_w_id ORDER BY o_w_id, o_d_id
+`)
+	if err != nil {
 		return err
 	}
 
-	// 3.3.2.3
-	// max(NO_O_ID) - min(NO_O_ID) + 1 = # of rows in new_order for each warehouse/district
-	var minNewOID, countNewOrder int
-	if err := forEachWarehouseAndDistrict(func(wID, dID int) error {
-		if err := db.QueryRow(`SELECT min(no_o_id), count(*) FROM new_order WHERE no_w_id=$1 AND no_d_id=$2`,
-			wID, dID,
-		).Scan(&minNewOID, &countNewOrder); err != nil {
+	var district, newOrder, order float64
+	var i int
+	for ; districtRows.Next() && newOrderRows.Next() && orderRows.Next(); i++ {
+		if err := districtRows.Scan(&district); err != nil {
 			return err
 		}
-		maxNewOID := maxNewOIDs[sliceIdx(wID, dID)]
-		if maxNewOID-minNewOID+1 != countNewOrder {
-			fmt.Printf("check failed: max(no_o_id)=%d + min(no_o_id)=%d + 1 != count(*)=%d for w_id=%d d_id=%d\n",
-				maxNewOID, minNewOID, countNewOrder, wID, dID)
+		if err := newOrderRows.Scan(&newOrder); err != nil {
+			return err
 		}
-		return nil
-	}); err != nil {
+		if err := orderRows.Scan(&order); err != nil {
+			return err
+		}
+
+		if (order != newOrder) || (order != (district - 1)) {
+			return errors.Errorf("inequality at idx %d: order: %d, newOrder: %d, district-1: %d",
+				i, order, newOrder, district-1)
+		}
+	}
+	if districtRows.Next() || newOrderRows.Next() || orderRows.Next() {
+		return errors.New("length mismatch between rows")
+	}
+	if err := districtRows.Close(); err != nil {
+		return err
+	}
+	if err := newOrderRows.Close(); err != nil {
+		return err
+	}
+	if err := orderRows.Close(); err != nil {
 		return err
 	}
 
-	// 3.3.2.4
-	// sum(O_OL_CNT) = # of rows in the ORDER-LINE table for each warehouse/district
-	var sumOOLCnt, countNewOrderLine int
-	if err := forEachWarehouseAndDistrict(func(wID, dID int) error {
-		if err := db.QueryRow(`SELECT sum(o_ol_cnt) FROM "order" WHERE o_w_id=$1 AND o_d_id=$2`,
-			wID, dID,
-		).Scan(&sumOOLCnt); err != nil {
-			return err
-		}
-		if err := db.QueryRow(`SELECT count(*) FROM order_line WHERE ol_w_id=$1 AND ol_d_id=$2`,
-			wID, dID,
-		).Scan(&countNewOrderLine); err != nil {
-			return err
-		}
-		if countNewOrderLine != sumOOLCnt {
-			fmt.Printf("check failed: count(order_line)=%d != sum(o_ol_cnt)=%d for w_id=%d d_id=%d\n",
-				countNewOrderLine, sumOOLCnt, wID, dID)
-		}
-		return nil
-	}); err != nil {
-		return err
-	}
-
-	// 3.3.2.8
-	// W_YTD = sum(H_AMOUNT) for each warehouse
-	var sumHAmount float64
-	for i := 0; i < *warehouses; i++ {
-		if err := db.QueryRow("SELECT w_ytd FROM warehouse WHERE w_id=$1", i).Scan(&wYTD); err != nil {
-			return err
-		}
-		if err := db.QueryRow("SELECT SUM(h_amount) FROM history WHERE h_w_id=$1", i).Scan(&sumHAmount); err != nil {
-			return err
-		}
-		if wYTD != sumHAmount {
-			fmt.Printf("check failed: w_ytd=%f != sum(h_amount)=%f for warehouse %d\n", wYTD, sumHAmount, i)
-		}
+	if i == 0 {
+		return errors.Errorf("zero rows")
 	}
 
 	return nil
+}
+
+func check3323(db *sql.DB) error {
+	// max(NO_O_ID) - min(NO_O_ID) + 1 = # of rows in new_order for each warehouse/district
+	rows, err := db.Query(`
+SELECT (VALUES (1)) FULL OUTER JOIN
+(SELECT MAX(no_o_id) - MIN(no_o_id) - COUNT(*) AS nod FROM new_order GROUP BY no_w_id, no_d_id)
+ WHERE nod != one
+`)
+	if err != nil {
+		return err
+	}
+
+	var i int
+	var val int
+	for ; rows.Next(); i++ {
+		if err := rows.Scan(&val); err != nil {
+			return err
+		}
+		if val != -1 {
+			return errors.Errorf("MAX(no_o_id) - MIN(no_o_id) - COUNT(rows) must be -1, got %d", val)
+		}
+	}
+
+	if err := rows.Close(); err != nil {
+		return err
+	}
+
+	if i == 0 {
+		return errors.Errorf("zero rows")
+	}
+
+	return nil
+}
+
+func check3324(db *sql.DB) error {
+	// SUM(O_OL_CNT) = [number of rows in the ORDER-LINE table for this district]
+
+	leftRows, err := db.Query(`
+SELECT SUM(o_ol_cnt) FROM tpcc.order GROUP BY o_w_id, o_d_id ORDER BY o_w_id, o_d_id
+`)
+	if err != nil {
+		return err
+	}
+	rightRows, err := db.Query(`
+SELECT COUNT(*) FROM tpcc.order_line GROUP BY ol_w_id, ol_d_id ORDER BY ol_w_id, ol_d_id
+`)
+	if err != nil {
+		return err
+	}
+	var i int
+	var left, right int64
+	for ; leftRows.Next() && rightRows.Next(); i++ {
+		if err := leftRows.Scan(&left); err != nil {
+			return err
+		}
+		if err := rightRows.Scan(&right); err != nil {
+			return err
+		}
+		if left != right {
+			return errors.Errorf("order.SUM(o_ol_cnt): %d != order_line.count(*): %d", left, right)
+		}
+	}
+	if i == 0 {
+		return errors.Errorf("0 rows returned")
+	}
+	if leftRows.Next() || rightRows.Next() {
+		return errors.Errorf("length of order.SUM(o_ol_cnt) != order_line.count(*)")
+	}
+
+	if err := leftRows.Close(); err != nil {
+		return err
+	}
+	return rightRows.Close()
+}
+
+func check3325(db *sql.DB) error {
+	// We want the symmetric difference between the sets:
+	// (SELECT no_w_id, no_d_id, no_o_id FROM tpcc.new_order)
+	// (SELECT o_w_id, o_d_id, o_id FROM tpcc.order@primary WHERE o_carrier_id IS NULL)
+	// We achieve this by two EXCEPT ALL queries.
+
+	firstQuery, err := db.Query(`
+(SELECT no_w_id, no_d_id, no_o_id FROM tpcc.new_order) 
+EXCEPT ALL 
+(SELECT o_w_id, o_d_id, o_id FROM tpcc.order@primary WHERE o_carrier_id IS NULL)`)
+	if err != nil {
+		return err
+	}
+	secondQuery, err := db.Query(`
+(SELECT o_w_id, o_d_id, o_id FROM tpcc.order@primary WHERE o_carrier_id IS NULL) 
+EXCEPT ALL 
+(SELECT no_w_id, no_d_id, no_o_id FROM tpcc.new_order)`)
+	if err != nil {
+		return err
+	}
+
+	if firstQuery.Next() {
+		return errors.Errorf("left EXCEPT right returned nonzero results.")
+	}
+
+	if secondQuery.Next() {
+		return errors.Errorf("right EXCEPT left returned nonzero results.")
+	}
+
+	if err := firstQuery.Close(); err != nil {
+		return err
+	}
+	return secondQuery.Close()
+}
+
+func check3326(db *sql.DB) error {
+	firstQuery, err := db.Query(`
+(SELECT o_w_id, o_d_id, o_id, o_ol_cnt FROM tpcc.order 
+  ORDER BY o_w_id, o_d_id, o_id DESC) 
+EXCEPT ALL 
+(SELECT ol_w_id, ol_d_id, ol_o_id, COUNT(*) FROM tpcc.order_line 
+  GROUP BY (ol_w_id, ol_d_id, ol_o_id) 
+  ORDER BY ol_w_id, ol_d_id, ol_o_id DESC)`)
+	if err != nil {
+		return err
+	}
+	secondQuery, err := db.Query(`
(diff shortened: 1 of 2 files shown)


tpcc/check.go
```diff
@@ -3,128 +3,270 @@ package main
...
+		return err
+	}
+
+	if rowsRight, err = db.Query("SELECT SUM(d_ytd) FROM district GROUP BY d_w_id ORDER BY d_w_id"); err != nil {
+		return err
+	}
+
+	var left, right float64
+	var i int
+	for ; rowsLeft.Next() && rowsRight.Next(); i++ {
```
petermattis: A client side merge join? Couldn't you combine the queries above into a FULL OUTER JOIN and then check whether the results had missing rows on either the left or right side?

tpcc/check.go
```diff
@@ -3,128 +3,270 @@ package main
...
+		rowsLeft.Scan(&left)
+		rowsRight.Scan(&right)
+
+		if left != right {
```
petermattis: Performing this via check via a join would also remove the possibility that a warehouse is missing from one of the tables.

jordanlewis:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[tpcc/check.go, line 29 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76XAcahN4YRQPtDIiA:-L76XAcahN4YRQPtDIiB:b-mhk3xy) ([raw file](https://github.com/cockroachdb/loadgen/blob/da5cf97a203d87edf0c271ff3cad8b3c95402a6b/tpcc/check.go#L29)):*
> ```Go
> 
> 	// 3.3.2.1
> 	// W_YTD = sum(D_YTD) for all warehouses and districts
> ```

I would prefer that you not delete these comments. They are helpful for understanding the intent of the check.

---

*[tpcc/check.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76bYWUGXkpv14a5nLg:-L76bYWUGXkpv14a5nLh:bf5uxf5) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L32)):*
> ```Go
> 
> 		if left != right {
> 			return errors.Errorf("warehouse %d: warehouse.w_ytd %f != sum(district.d_ytd) %f",
> ```

the previous version had the good (imo) property of not quitting on error. That way you can see all of the failures at once. I'd prefer to preserve that if possible

---

*[tpcc/check.go, line 209 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76cWqhpo4nVn9aDpJj:-L76cWqhpo4nVn9aDpJk:b-h1q4ym) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L209)):*
> ```Go
> 
> func check3326(db *sql.DB) error {
> 	// return errors.Errorf("Do not call this check, as its prohibitvely expensive until CockroachDB has a streaming GROUP BY aggregation.")
> ```

is this still true? maybe you should add an `-expensive` flag that controls whether it's run.

---

*[tpcc/check.go, line 213 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76d2FQ6hqwIGng5zSB:-L76d2FRxVQ6VP2biL76:b-x1kn1u) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L213)):*
> ```Go
> 
> 	// We want the symmetric difference between the sets:
> 	// (SELECT o_w_id, o_d_id, o_id, o_ol_cnt FROM tpcc.order ORDER BY o_w_id, o_d_id, o_id DESC)
> ```

Should these comments go byebye?

---

*[tpcc/check.go, line 241 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76Wu9sG7ETinDIhhHN:-L76Wu9sG7ETinDIhhHO:b-bti7rm) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L241)):*
> ```Go
> 	}
> 
> 	fmt.Printf("3.3.2.6: passed in %f seconds.\n", time.Since(start).Seconds())
> ```

These prints and associated timing code are duplicated and should be done just once in the loop. You should change `checkFunc` to be a struct that contains both the name of the check and the check function itself to facilitate this.

---

*[tpcc/check.go, line 251 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76YEAPID3eW5wFpu5R:-L76YEAPID3eW5wFpu5S:bc0ua4p) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L251)):*
> ```Go
> 	// single transaction as otherwise we cannot run these checks
> 	// on a live running TPC-C database.
> 	db.SetMaxOpenConns(10)
> ```

This doesn't do anything wrong, but I'm curious why you added this. I don't believe there is any concurrency to the queries in this routine.

---

*[tpcc/check.go, line 257 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76XzewDNnVhMUO-LuI:-L76XzewDNnVhMUO-LuJ:buht6u3) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L257)):*
> ```Go
> 	}
> 	expensiveChecks := []checkFunc{
> 		check3326,
> ```

So far AFAICT there's no point in having these separated. Just collapse them. If you take my earlier suggestion about making `checkFunc` into a struct with a name, you can add a bool `expensive` to that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

tpcc/check.go
```diff
@@ -3,129 +3,274 @@ package main
...
-		}
-		if err := db.QueryRow(`SELECT max(no_o_id) FROM new_order WHERE no_w_id=$1 AND no_d_id=$2`,
-			wID, dID,
-		).Scan(&maxNewOID); err != nil {
-			return err
-		}
-		maxNewOIDs[sliceIdx(wID, dID)] = maxNewOID
+	var district, newOrder, order float64
+	var i int
+	for ; districtRows.Next() && newOrderRows.Next() && orderRows.Next(); i++ {
```
petermattis: A 3-way client side merge join? You must have a reason for doing this vs a more complex query.

tpcc/check.go
```diff
@@ -3,129 +3,274 @@ package main
...
+}
+
+func check3324(db *sql.DB) error {
+	start := time.Now()
+
+	leftRows, err := db.Query("SELECT SUM(o_ol_cnt) FROM tpcc.order GROUP BY o_w_id, o_d_id ORDER BY o_w_id, o_d_id")
+	if err != nil {
+		return err
+	}
+	rightRows, err := db.Query("SELECT count(*) from tpcc.order_line GROUP BY ol_w_id, ol_d_id ORDER BY ol_w_id, ol_d_id")
```
petermattis: Nit: `s/count/COUNT/g`

petermattis: `s/from/FROM/g`

jordanlewis:
The actual check logic :lgtm: though!

---

Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-:-L76duilWD0B62U7nrIU:b-rb2qxi)*

tpcc/check.go
```diff
@@ -3,129 +3,274 @@ package main
...
+
+func check3325(db *sql.DB) error {
+	start := time.Now()
+
+	// We want the symmetric difference between the sets:
+	// (SELECT no_w_id, no_d_id, no_o_id FROM tpcc.new_order)
+	// (SELECT o_w_id, o_d_id, o_id FROM tpcc.order@primary WHERE o_carrier_id IS NULL)
+	// We achieve this by two EXCEPT ALL queries.
+
+	firstQuery, err := db.Query("(SELECT no_w_id, no_d_id, no_o_id FROM tpcc.new_order) EXCEPT ALL (SELECT o_w_id, o_d_id, o_id FROM tpcc.order@primary WHERE o_carrier_id IS NULL)")
```
petermattis: Nit: you can add newlines to these queries to make them more readable. For example:

```
`(SELECT no_w_id, no_d_id, no_o_id FROM tpcc.new_order) 
EXCEPT ALL 
(SELECT o_w_id, o_d_id, o_id FROM tpcc.order@primary WHERE o_carrier_id IS NULL)`
```

(Note that this is a back-quoted string)

petermattis (approved):

rjnn:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[tpcc/check.go, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r1-27:-L7BMnkMn6xN4padqfVz:b2ada79) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L27)):*

A client side merge join? Couldn't you combine the queries above into a FULL OUTER JOIN and then check whether the results had missing rows on either the left or right side?
</blockquote></details>

I stopped optimizing this because we reduced the check time to 0.05 seconds for 100 warehouses, and I had other checks to work on, but sure. Done. The results are... not pretty.

---

*[tpcc/check.go, line 29 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76XAcahN4YRQPtDIiA:-L7BiYutdSFwEiURA-Y3:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/da5cf97a203d87edf0c271ff3cad8b3c95402a6b/tpcc/check.go#L29)):*

I would prefer that you not delete these comments. They are helpful for understanding the intent of the check.
</blockquote></details>

Done.

---

*[tpcc/check.go, line 31 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r1-31:-L7BiHn6-QcG5qMzT-Zx:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L31)):*

Performing this via check via a join would also remove the possibility that a warehouse is missing from one of the tables.
</blockquote></details>

Done.

---

*[tpcc/check.go, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76bYWUGXkpv14a5nLg:-L7AcH6QuJeEwdr_eXt6:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L32)):*

the previous version had the good (imo) property of not quitting on error. That way you can see all of the failures at once. I'd prefer to preserve that if possible
</blockquote></details>

Done.

---

*[tpcc/check.go, line 209 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76cWqhpo4nVn9aDpJj:-L7Bs6QcvweZVGXSPkdG:bk6cnkv) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L209)):*

is this still true? maybe you should add an `-expensive` flag that controls whether it's run.
</blockquote></details>

Yes. We need streaming GROUP BY for this to be fast. 

`-expensive` flag is done.

---

*[tpcc/check.go, line 213 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76d2FQ6hqwIGng5zSB:-L7Ad8qQrauy7SzeOqo0:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L213)):*

Should these comments go byebye?
</blockquote></details>

Done.

---

*[tpcc/check.go, line 241 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76Wu9sG7ETinDIhhHN:-L7BLSZNZj7A-pfg3-n8:b7qhmqu) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L241)):*

These prints and associated timing code are duplicated and should be done just once in the loop. You should change `checkFunc` to be a struct that contains both the name of the check and the check function itself to facilitate this.
</blockquote></details>

Done. So much more elegant!

---

*[tpcc/check.go, line 251 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76YEAPID3eW5wFpu5R:-L7BMFhyrm0ZsegjH2f5:b-8buf0o) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L251)):*

This doesn't do anything wrong, but I'm curious why you added this. I don't believe there is any concurrency to the queries in this routine.
</blockquote></details>

I need multiple connections as I have multiple open `Rows` at a time. Added a comment and TODO to clarify.

---

*[tpcc/check.go, line 257 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76XzewDNnVhMUO-LuI:-L7BMhcON07htfC40-6v:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L257)):*

So far AFAICT there's no point in having these separated. Just collapse them. If you take my earlier suggestion about making `checkFunc` into a struct with a name, you can add a bool `expensive` to that.
</blockquote></details>

Done.

---

*[tpcc/check.go, line 73 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r2-73:-L7BrAopWZLF8nPXUL8f:b-hozm82) ([raw file](https://github.com/cockroachdb/loadgen/blob/2bbfc9a2f18c3c6fbdd996dc862412d34857b082/tpcc/check.go#L73)):*

A 3-way client side merge join? You must have a reason for doing this vs a more complex query.
</blockquote></details>

It's _way_ faster than any alternative I can think of, and only requires 3 queries that return O(warehouses) rows.

`SELECT * FROM (SELECT MAX(no_o_id) FROM tpcc.new_order GROUP BY no_d_id, no_w_id) AS newordermax JOIN (SELECT d_next_o_id FROM district) ON newordermax.max = (d_next_o_id - 1) JOIN (SELECT MAX(o_id) FROM tpcc.order GROUP BY o_d_id, o_w_id) as ordermax on ordermax.max = newordermax.max` took me a minute to compute on 100 warehouses, and is unreadable.

I'm also only optimizing these queries until they are fast enough to run, yet still readable, and no further.

---

*[tpcc/check.go, line 141 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r2-141:-L7AcQ8f_bojWnJ8NXC2:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/2bbfc9a2f18c3c6fbdd996dc862412d34857b082/tpcc/check.go#L141)):*

`s/from/FROM/g`
</blockquote></details>

Done.

---

*[tpcc/check.go, line 180 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r2-180:-L7AcmHGOZ8gzMarBkpM:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/2bbfc9a2f18c3c6fbdd996dc862412d34857b082/tpcc/check.go#L180)):*

Nit: you can add newlines to these queries to make them more readable. For example:

```
`(SELECT no_w_id, no_d_id, no_o_id FROM tpcc.new_order) 
EXCEPT ALL 
(SELECT o_w_id, o_d_id, o_id FROM tpcc.order@primary WHERE o_carrier_id IS NULL)`
```

(Note that this is a back-quoted string)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

rjnn:
*[tpcc/check.go, line 73 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r2-73:-L7Bt78FEcNtB07xGr-g:btw5yxf) ([raw file](https://github.com/cockroachdb/loadgen/blob/2bbfc9a2f18c3c6fbdd996dc862412d34857b082/tpcc/check.go#L73)):*

It's _way_ faster than any alternative I can think of, and only requires 3 queries that return O(warehouses) rows.

`SELECT * FROM (SELECT MAX(no_o_id) FROM tpcc.new_order GROUP BY no_d_id, no_w_id) AS newordermax JOIN (SELECT d_next_o_id FROM district) ON newordermax.max = (d_next_o_id - 1) JOIN (SELECT MAX(o_id) FROM tpcc.order GROUP BY o_d_id, o_w_id) as ordermax on ordermax.max = newordermax.max` took me a minute to compute on 100 warehouses, and is unreadable.

I'm also only optimizing these queries until they are fast enough to run, yet still readable, and no further.
</blockquote></details>

Update: i was wrong, it did not take me a minute, instead it crashed my gateway node.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

jordanlewis:
:lgtm:

---

Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-:-L7BtkXAUZuP5CbUIzMm:bnfp4nl)*

petermattis:
:lgtm:

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[tpcc/check.go, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r1-27:-L7CN3UVG761c-rcOQDX:b-5ixuc) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L27)):*

I stopped optimizing this because we reduced the check time to 0.05 seconds for 100 warehouses, and I had other checks to work on, but sure. Done. The results are... not pretty.
</blockquote></details>

Hmm, I thought this would turn into something like:

```
SELECT w_id, w_ytd, sum_d_ytd 
FROM warehouse 
FULL OUTER JOIN 
(SELECT d_w_id, SUM(d_ytd) as sum_d_ytd FROM district GROUP BY d_w_id) 
ON (w_id = d_w_id)
WHERE w_ytd != sum_d_ytd
```

I might have done that wrong, but the idea is that you'll get rows whenever one side contains a warehouse not in the other, or when `w_ytd != sum_d_ytd`. 

---

*[tpcc/check.go, line 180 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r2-180:-L7COiLxs1uwCmEW5tJO:b-a20mgy) ([raw file](https://github.com/cockroachdb/loadgen/blob/2bbfc9a2f18c3c6fbdd996dc862412d34857b082/tpcc/check.go#L180)):*

Done.
</blockquote></details>

There is at least one query above that deserves this treatment as well (it exceeds 100 columns in length).

---

*[tpcc/check.go, line 238 at r3](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7COpI5BjIs2ylnsmQY:-L7COpI5BjIs2ylnsmQZ:b-gjgn83) ([raw file](https://github.com/cockroachdb/loadgen/blob/7b23353b3dc9b4a2e5a4132d5795ec05acc75948/tpcc/check.go#L238)):*
> ```Go
> func checkConsistency(db *sql.DB) bool {
> 	// TODO(arjun): each query should only require a single connection at a time, this should not be required.
> 	db.SetMaxOpenConns(3)
> ```

I think you can just remove this line completely. The default is to allow as many connections as necessary, which is fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-:-L7COynMYCv2wjIvNG12:bnfp4nl)*

rjnn:
@jordanlewis do you know what's up with CI?

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[tpcc/check.go, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r1-27:-L7CUByytUXziVwY_AIU:b-dcjnl5) ([raw file](https://github.com/cockroachdb/loadgen/blob/fde5b82929e881187d43cbf4c43e1d07848222b3/tpcc/check.go#L27)):*

Hmm, I thought this would turn into something like:

```
SELECT w_id, w_ytd, sum_d_ytd 
FROM warehouse 
FULL OUTER JOIN 
(SELECT d_w_id, SUM(d_ytd) as sum_d_ytd FROM district GROUP BY d_w_id) 
ON (w_id = d_w_id)
WHERE w_ytd != sum_d_ytd
```

I might have done that wrong, but the idea is that you'll get rows whenever one side contains a warehouse not in the other, or when `w_ytd != sum_d_ytd`. 
</blockquote></details>

Your formulation is much more elegant. Going with it. Thanks!

---

*[tpcc/check.go, line 180 at r2](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L76VixjofXdwmnDJ8W5-r2-180:-L7CYBA98SiH4x-mCrBS:b-896fix) ([raw file](https://github.com/cockroachdb/loadgen/blob/2bbfc9a2f18c3c6fbdd996dc862412d34857b082/tpcc/check.go#L180)):*

There is at least one query above that deserves this treatment as well (it exceeds 100 columns in length).
</blockquote></details>

Done.

---

*[tpcc/check.go, line 238 at r3](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7COpI5BjIs2ylnsmQY:-L7CUHkFgpTXnTH1bJIw:b-jbrsat) ([raw file](https://github.com/cockroachdb/loadgen/blob/7b23353b3dc9b4a2e5a4132d5795ec05acc75948/tpcc/check.go#L238)):*

I think you can just remove this line completely. The default is to allow as many connections as necessary, which is fine.
</blockquote></details>

The default when I was running them was one. This is because this is set by `setupDatabase` to `*warehouses/len(dbURLs)`, and the default for `*warehouses` is 1.  I wasn't specifying the warehouses flag, as why would you, when running checks? The checks are just the checks.

I have removed this line and tweaked the `setupDatabase` to not set a `maxOpenConnections` if `*checks` is true.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-:-L7CTU_44o4r5dB2PW0O:bvumbhm)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[tpcc/check.go, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7CYsLLmxN4ULpSFqmT:-L7CYsLLmxN4ULpSFqmU:btq0a1c) ([raw file](https://github.com/cockroachdb/loadgen/blob/51128120d0d32c9e3a8018abf873b53db7bdc544/tpcc/check.go#L32)):*
> ```Go
> 	var i int
> 	for ; rows.Next(); i++ {
> 		rows.Scan(&val)
> ```

You're not using this value, is there any reason you're scanning it? I think you could gather up a list of warehouse IDs here. Or you could change the query to do a `SELECT COUNT(*)` instead of `SELECT w_id, w_ytd, sum_d_ytd`. Use all the SQL!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

rjnn:
*[tpcc/check.go, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7CYsLLmxN4ULpSFqmT:-L7CaiFTy_5stqbRoPKu:b-kn2yj5) ([raw file](https://github.com/cockroachdb/loadgen/blob/51128120d0d32c9e3a8018abf873b53db7bdc544/tpcc/check.go#L32)):*

You're not using this value, is there any reason you're scanning it? I think you could gather up a list of warehouse IDs here. Or you could change the query to do a `SELECT COUNT(*)` instead of `SELECT w_id, w_ytd, sum_d_ytd`. Use all the SQL!
</blockquote></details>

Done. This impromptu SQL quiz has rapidly turned embarrassing!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[tpcc/check.go, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7CYsLLmxN4ULpSFqmT:-L7Cb2ojG2BO2--PM9iM:bubi86z) ([raw file](https://github.com/cockroachdb/loadgen/blob/51128120d0d32c9e3a8018abf873b53db7bdc544/tpcc/check.go#L32)):*

Done. This impromptu SQL quiz has rapidly turned embarrassing!
</blockquote></details>

Heh, thinking about the optimizer has massively expanded my SQL repertoire. 

Nit: `row.Scan()` returns an `error`. You should check it, otherwise the query can fail silently and `i` will be equal to `0`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

rjnn:
*[tpcc/check.go, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7CYsLLmxN4ULpSFqmT:-L7CbZZqCtSyfyxehvih:b-gbx872) ([raw file](https://github.com/cockroachdb/loadgen/blob/51128120d0d32c9e3a8018abf873b53db7bdc544/tpcc/check.go#L32)):*

Heh, thinking about the optimizer has massively expanded my SQL repertoire. 

Nit: `row.Scan()` returns an `error`. You should check it, otherwise the query can fail silently and `i` will be equal to `0`.
</blockquote></details>

Done. This whole thing has been a pretty fun SQL mini-exercise as well, I must admit!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[tpcc/check.go, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7CYsLLmxN4ULpSFqmT:-L7Ccs2cEAljGVF69h6Y:b-xr9eiq) ([raw file](https://github.com/cockroachdb/loadgen/blob/51128120d0d32c9e3a8018abf873b53db7bdc544/tpcc/check.go#L32)):*

Done. This whole thing has been a pretty fun SQL mini-exercise as well, I must admit!
</blockquote></details>

:shipit:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

rjnn:
*[tpcc/check.go, line 32 at r4](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-L7CYsLLmxN4ULpSFqmT:-L7Cd9_8-3jcg1ANn53o:b-wui0fe) ([raw file](https://github.com/cockroachdb/loadgen/blob/51128120d0d32c9e3a8018abf873b53db7bdc544/tpcc/check.go#L32)):*

:shipit:
</blockquote></details>

I can't until we fix this broken CI pipeline, so I'm just going to keep adding to this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161)*

rjnn:
Turns out the CI failures are that you just have to click through a bunch of teamcity buttons to get the answer: `make check` failed. Fixed the style checks.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/161#-:-L7Cg_KyT-xHoPngSrRY:b-dfjg3u)*

PR #166, opened 2018-03-12
tpcc: allow specifying which checks to run

### LINKED ISSUE #164: tpcc: run subset of checks
It takes over 10 minutes to run check 3.3.2.4 on 10k warehouses. Some other checks failed to run because I closed the connection of the machine running the checks by accident. It would be great to be able to run just some of the checks instead of all of them.

rjnn:
Also removed the -expensive flag, since now the user can make that
determination on their own.

Closes #164.


tpcc/main.go
```diff
@@ -37,7 +37,7 @@ var check = flag.Bool("check", false, "Run consistency checks.")
 var concurrency = flag.Int("concurrency", 2*runtime.NumCPU(), "Number of terminals (ignored unless -no-wait is specified)")
 var drop = flag.Bool("drop", false, "Drop the database and recreate")
 var duration = flag.Duration("duration", 0, "The duration to run. If 0, run forever.")
-var expensive = flag.Bool("expensive", false, "Run expensive checks. (ignored unless -check is specified)")
+var checksToRun = flag.String("only-checks", "1,2,3,4,5,6", "Which consistency checks to run (ignored unless -check is specified).")
```
jordanlewis: Rather than hard coding this to the list of current available checks, you should make the default empty and run all the checks if this flag is empty.

tpcc/check.go
```diff
@@ -227,28 +231,33 @@ EXCEPT ALL
...
+	toRun := strings.Split(*checksToRun, ",")
+	for _, checkToRun := range toRun {
+		i, err := strconv.Atoi(checkToRun)
+		if err != nil {
+			fmt.Printf("error parsing check %s, skipping: %v", checkToRun, err)
+			errorEncountered = true
 			continue
 		}
+
+		check := checks[i-1]
```
jordanlewis: Not a big deal because this is run manually, but this will panic if you specify a check that doesn't exist. I suggest creating a map from int check id to check before this loop by looping through the checks, and then using the map in this loop instead of the slice. Then you can throw an error if the check is not found.

jordanlewis (commented):
Thank you for the quick turnaround on this!

PR #167, opened 2018-03-13
tpcc: partition workers across connection pools

### LINKED ISSUE #157: tpcc: assign workers to connections based on partitions
Now that the tables are partitioned, we should assign workers to DB connection pools based on the partitions. At the moment, workers are assigned to connection pools by doing:

```
	for i := range workers {
		w := i % *warehouses
		workers[i] = newWorker(i, w, dbs[w%len(dbs)], &wg)
	}
```

So warehouse `w` is assigned to DB connection pool `w%numPools`. This is at odds with the range-based partitioning of the tables by warehouse ID. Contiguous warehouse IDs will be on the same partition. The above initialization should be fixed so that we determine the partition for a warehouse and from that partition we round-robin amongst the connection pools that are connected to the partition.

petermattis:
More aggressive splitting of the district, warehouse and history tables
which is useful for tpcc-10k to get better balancing of these tables.

Fixes #157

### DIFF
--- a/tpcc/ddls.go
+++ b/tpcc/ddls.go
@@ -268,8 +268,8 @@ func loadSchema(db *sql.DB, interleave bool, index bool, usePostgres bool) {
 // NB: Since we always split at the same points (specific warehouse IDs and
 // item IDs), splitting is idempotent.
 func splitTables(db *sql.DB, warehouses int) {
-	// Split district and warehouse tables every 100 warehouses.
-	const warehousesPerRange = 100
+	// Split district and warehouse tables every 10 warehouses.
+	const warehousesPerRange = 10
 	for i := warehousesPerRange; i < warehouses; i += warehousesPerRange {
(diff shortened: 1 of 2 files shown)


petermattis:
Note that to get the partitioning of connections, you have to specify the same `-partitions` value when running as the number of partitions. The default behavior (`-partions=0`) results in the old assignment of workers to connections.

jordanlewis:
I don't get it. What is the new connection partitioning scheme exactly, and how does it differ from the old scheme? Was the problem that we were arbitrarily assigning warehouse to node, instead of corresponding to which partitions a node owns?

petermattis:
> Was the problem that we were arbitrarily assigning warehouse to node, instead of corresponding to which partitions a node owns?

Yes. Warehouses are mapped to partitions and the mapping of workers to nodes was not matching that layout. The upside of this change is that we see more locality of access and thus more local RPCs and fewer remote RPCs.

jordanlewis (approved):

Issue #151, opened 2018-03-03
tpcc: sanity check txn requirements


petermattis:
The TPC-C spec has sanity checks on the randomized txns (9.2.2.5):

1. At least 0.9% and at most 1.1% of the New-Order transactions must roll back as a result of an unused item number.
2. The average number of order-lines per order must be in the range of 9.5 to 10.5 and the number of order- lines per order must be uniformly distributed from 5 to 15 for the New-Order transactions that are submitted to the SUT during the measurement interval.
3. The number of remote order-lines must be at least 0.95% and at most 1.05% of the number of order-lines that are filled in by the New-Order transactions that are submitted to the SUT during the measurement interval.
4. The number of remote Payment transactions must be at least 14% and at most 16% of the number of Payment transactions that are submitted to the SUT during the measurement interval.
5. The number of customer selections by customer last name in the Payment transaction must be at least 57% and at most 63% of the number of Payment transactions that are submitted to the SUT during the measurement interval.
6. The number of customer selections by customer last name in the Order-Status transaction must be at least 57% and at most 63% of the number of Order-Status transactions that are submitted to the SUT during the measurement interval.

We should add metrics to track the above and output these metrics when the test ends along with an indication of whether we met the requirements.

petermattis:
@arjunravinarayan This is related to fleshing out the consistency checks and would be another level of sanity checking that we should perform.

rjnn:
SGTM.

rjnn:
7. Verify that no more than 1%, or no more than one (1), whichever is greater, of the Delivery transactions skipped because there were fewer than necessary orders present in the New-Order table. (9.2.1.7)
8. `O_ALL_LOCAL` can be used to verify that approximately 10% of all orders contain at least one remote order-line. (9.2.2.9)
9. Perform distributional checks on `S_YTD`, `S_ORDER_CNT`, `S_REMOTE_CNT`, `C_PAYMENT_CNT` (also 9.2.2.9).

solongordon:
This is all set except for the 9.2.2.9 checks. I propose not adding those because a) they're kind of vague and b) the TPC-C spec makes them sound optional ("The auditor can further verify the compliance..."). Any objections?

solongordon:
Relevant PRs for future reference:
- https://github.com/cockroachdb/cockroach/pull/24177
- https://github.com/cockroachdb/cockroach/pull/24219
- https://github.com/cockroachdb/cockroach/pull/24230
[...]

PR #173, opened 2018-03-21
tpcc: add auditor and initial check

solongordon:
I added some basic auditing functionality to the TPC-C workload. This is
similar to the existing checks in checks.go, but rather than querying
the db it relies on statistics which are gathered as the workload runs.
Right now the audit checks are performed only when the workload exits,
but in the future we could also consider running them periodically as it
runs.

Refers #151

### DIFF
--- a/tpcc/audit.go
+++ b/tpcc/audit.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"fmt"
+	"sync/atomic"
+
+	"github.com/pkg/errors"
+)
+
+// auditor maintains statistics about TPC-C input data and runs distribution
+// checks, as specified in Clause 9.2 of the TPC-C spec.
+type auditor struct {
+	newOrderTransactions uint64
+	newOrderRollbacks    uint64
+}
+
+// runChecks runs the audit checks and prints warnings to stdout for those that
+// fail.
+func (a *auditor) runChecks() {
+	type check struct {
+		name string
+		f    func(a *auditor) error
+	}
+	checks := []check{
+		{"9.2.2.5.1", check92251},
+	}
+	for _, check := range checks {
+		result := check.f(a)
+		if result != nil {
+			fmt.Println(errors.Wrapf(result, "WARN: Failed audit check %s", check.name))
+		}
+	}
+}
+
+func check92251(a *auditor) error {
+	// At least 0.9% and at most 1.1% of the New-Order transactions roll back as a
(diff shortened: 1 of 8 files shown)


rjnn:
:lgtm: Great start!

---

Reviewed 8 of 8 files at r1.

---

*[tpcc/audit.go, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/173#-L88zjJdT0Ozb6gtNyKo:-L88zjJdT0Ozb6gtNyKp:b-9xlch2) ([raw file](https://github.com/cockroachdb/loadgen/blob/e3c9b2cce5e1c996a6a3af5184a18919251f0b0c/tpcc/audit.go#L39)):*
> ```Go
> 	// result of an unused item number.
> 	orders := atomic.LoadUint64(&a.newOrderTransactions)
> 	if orders < 0 {
> ```

less than zero?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/173#-:-L89-9aK65RFTezcwf9B:b-kt083j)*

solongordon:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[tpcc/audit.go, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/loadgen/173#-L88zjJdT0Ozb6gtNyKo:-L89-hg-BhF-XuXkaRGA:b-8rhlwa) ([raw file](https://github.com/cockroachdb/loadgen/blob/e3c9b2cce5e1c996a6a3af5184a18919251f0b0c/tpcc/audit.go#L39)):*

less than zero?
</blockquote></details>

Oops good catch. :) Switched this for testing and forgot to change it back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/173)*

rjnn:
:shipit: 

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/loadgen/173#-:-L89-zD76YWu3FxhQzqB:bizlbj6)*

jordanlewis:
The spec technically requires reporting all of these statistics at the end of the run even if we passed them, so it could be good to get these printed to stdout at the end. I'd be happy leaving that to a future PR though.

solongordon:
Noted, thanks @jordanlewis.

PR #175, opened 2018-03-21
tpcc: fix random range of several events

jordanlewis:
rand.Intn is exclusive on the right, not inclusive. This caused several
mistakes in implementation.

1. `11.11...%` of customers in the initial dataset have bad credit
instead of 10%. this causes the payment transaction to do slightly more
work - when customers have bad credit, some strings must be appended and
returned to the user.

2. the read-only orderStatus transaction was only selecting over 9
districts, instead of 10. this will slightly reduce contention when
fixed.

3. `11.11...%` of items have the string ORIGINAL in them instead of 10%.
this has no appreciable effect.

4. the read-only stockLevel transaction only selected over 9 districts,
instead of 10. this will slightly reduce contention when fixed.
```

### DIFF
--- a/tpcc/generate.go
+++ b/tpcc/generate.go
@@ -213,7 +213,7 @@ VALUES `
 					// 10% of the customer rows have bad credit.
 					// See section 4.3, under the CUSTOMER table population section.
 					credit := goodCredit
-					if rand.Intn(9) == 0 {
+					if rand.Intn(10) == 0 {
 						// Poor 10% :(
 						credit = badCredit
 					}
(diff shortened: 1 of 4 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/loadgen/175#-:-L89zZ7BNYonRRhh00Cq:bnfp4nl)*

PR #179, opened 2018-04-02
tpcc: fix max & idle connection property for each db.

louishust:


jordanlewis:
Hi @louishust, we've migrated development of this tool to `workload`, which lives in the main cockroach repo, so I'm going to close this PR. (It's also not a bug - the intent of this is to set a large number of connections only when we're not in the "check" mode).

We're going to push a deprecation announcement to this repo soon. But never fear, the code is still alive and well - you'll just run it via `workload` instead of this binary.

louishust:
hi @jordanlewis , is there some document for hot to use workload?

